PR #24, opened 2014-12-30
Thread Tile Writing and Speed Up Pbf Parsing

kevinkreiser:
This pr adds threaded writing of tiles. Currently its hardcoded to spawn 4 threads. In practice this doesnt seem to do a good job of writing in parallel. Often a thread will finish writing all of its tiles before the next thread starts up. Hopefully I can figure out why and fix that.

I noticed that the bulk of the time spent makign tiles though is spent in parsing the pbf, rather running over the in memory data structure constructed from the pbf file. I added a bit of code to tell the ripped off code that we have to skip the bits we dont need and this significantly sped up the PA import. With the default optimization flags this made a huge difference something like 2m 30s down to 1m 50s. With the -O3 flag turned on the threaded branch is only 7% faster or something like that. I think what we really need to do is make parsing of the pbf threaded and write the parsers ourselves (there is a lot of copying around of data in there).

### DIFF
--- a/src/mjolnir/graphtilebuilder.cc
+++ b/src/mjolnir/graphtilebuilder.cc
@@ -1,6 +1,7 @@
 #include "mjolnir/graphtilebuilder.h"
 
 #include <boost/filesystem/operations.hpp>
+#include <stdexcept>
 
 using namespace valhalla::baldr;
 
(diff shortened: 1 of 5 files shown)


kevinkreiser:
opening another pr based on this

PR #25, opened 2015-01-01
Thread Tile Writing and Speed Up Pbf Parsing

kevinkreiser:
This pr adds threaded writing of tiles. Currently its hardcoded to spawn 4 threads. In practice this doesnt seem to do a good job of writing in parallel. Often a thread will finish writing all of its tiles before the next thread starts up. After looking at this for a while I am wondering if the threads are suffering from: http://en.wikipedia.org/wiki/False_sharing or something similar. Basically the different threads are all fighting to get at the same objects from memory but different parts, this puts a real strain on the L 1-3 caches because the cpu can't figure out which threads stuff should go into local cache. i'll write a toy program to test this theory. i was really hoping that when one thread was writing to file another thread could be reading from memory and filling out the data structure. we'll see..

I noticed that the bulk of the time spent makign tiles though is spent in parsing the pbf, rather running over the in memory data structure constructed from the pbf file. I added a bit of code to tell the ripped off code that we have to skip the bits we dont need and this significantly sped up the PA import. With the default optimization flags this made a huge difference something like 2m 30s down to 1m 50s. With the -O3 flag turned on the threaded branch is only 7% faster or something like that. In total the whole things runs in just under a minute (usually 58 seconds). I think what we really need to do is make parsing of the pbf threaded and write the parsers ourselves (there is a lot of copying around of data in there).

This branch includes some other fixes I found along the way. There was a bug in the nodetable which I fixed via adding a test (this didn't require me to but was way easier to do once I pulled it out into the library). There were more memory leak fixes done as well.


src/mjolnir/graphbuilder.cc
```diff
@@ -0,0 +1,788 @@
...
+        edge.targetnode_ = currentid;
+        nd.AddEdge(edgeindex);
+
+        // Add the edge to the list of edges
+        edges_.push_back(edge);
+        edgeindex++;
+
+        // Start a new edge if this is not the last node in the way
+        if (i < n-1) {
+          edge = Edge(currentid, wayindex, node.latlng());
```
kevinkreiser: this doesnt make sense. why are we using node.latlng (ie the first point on the way) instead of nd.latlng? when i change it, everything gets messed up. can someone explain this to my belabored brain?

dnesbitt61:
ship

kevinkreiser:
thanks, will

PR #26, opened 2015-01-02
Add not_thru flags to directed edges

dnesbitt61:
Computes this when buildign directed edges.

@kevinkreiser @dgearhart @gknisely


src/mjolnir/pbfgraphbuilder/graphbuilder.cc
```diff
@@ -489,6 +495,69 @@ uint32_t GetOpposingIndex(const uint64_t endnode, const uint64_t startnode, cons
   return 31;
 }
 
+// Test if this is a "not thru" edge. These are edges that enter a region that
+// has no exit other than the edge entering the region
+bool IsNoThroughEdge(const uint64_t startnode, const uint64_t endnode,
+                     const uint32_t startedgeindex,
+                     const std::unordered_map<uint64_t, OSMNode>& nodes,
+                     const std::vector<Edge>& edges) {
+  std::set<uint64_t> visitedset;
```
kevinkreiser: if the order doesnt matter of these this would be faster with an underedered_set

src/mjolnir/pbfgraphbuilder/graphbuilder.cc
```diff
@@ -489,6 +495,69 @@ uint32_t GetOpposingIndex(const uint64_t endnode, const uint64_t startnode, cons
...
+
+        // TODO - if edges had the road class we could end search here if a
+        // major rado is encountered
+
+        // Add to the expand set
+        expandset.emplace(osmendnode);
+      }
+    }
+  }
+  visitedset.clear();
```
kevinkreiser: no need to do this as it will happen when leaving the function

src/mjolnir/pbfgraphbuilder/graphbuilder.cc
```diff
@@ -489,6 +495,69 @@ uint32_t GetOpposingIndex(const uint64_t endnode, const uint64_t startnode, cons
...
+      visitedset.clear();
+      expandset.clear();
+      return true;
+    }
+    n++;
+
+    // Get the node off of the expand list and add it to the visited list
+    node = *expandset.begin();
+    expandset.erase(expandset.begin());
+    visitedset.emplace(node);
```
kevinkreiser: as far as i can tell visitedset is never read from, whats the point of it?

kevinkreiser:
any chance we can get the other pr in mjolnir merged first so i dont have to manage the merge conflicts? :smile:

src/mjolnir/pbfgraphbuilder/graphbuilder.cc
dnesbitt61: thanks. I thought so, but then got worried about multi-threading so did that to be safe. Removing.

kevinkreiser: yeah the good thing about it being an anonymous namespace method and not a member one is that any thread can call it and the mem allocated there has nothing to do with any other thread of execution so its completely safe

kevinkreiser:
other than above, ship

PR #29, opened 2015-01-05
Fix bug in setting the 1st lat,lng of a new edge in ConstructEdges

dnesbitt61:


kevinkreiser:
:point_right: :bear:

PR #33, opened 2015-01-06
Add sorting of the edge indexes based on driveability and importance.

dnesbitt61:
@dgearhart @kevinkreiser @gknisely


gknisely:
ship it

PR #36, opened 2015-01-06
Slight optimization of not_thru processing

dnesbitt61:
Assume the road is thru if we hit a tertiary or better class road.

@dgearhart @gknisely @kevinkreiser


gknisely:
ship

PR #38, opened 2015-01-07
write compressed shape to tile

kevinkreiser:
this cuts the size of the tiles in half. enough said.

just kidding. ive also removed a bunch of unneeded code to hopefully simplfy the edgeinfobuilder object. added some more copy elision in the actual constructing of the tiles (this time for the edges). used ostream << operators to do writing where we can. changed the test for edgeinfobuilder (and edgeinfo) to use updated interface.

depends on: https://github.com/valhalla/baldr/pull/36

### DIFF
--- a/src/mjolnir/edgeinfobuilder.cc
+++ b/src/mjolnir/edgeinfobuilder.cc
@@ -2,102 +2,50 @@
 #include <ostream>
 #include <iostream>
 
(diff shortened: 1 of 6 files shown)


gknisely:
k

dnesbitt61:
whew...ok by me

dgearhart:
:+1:

PR #40, opened 2015-01-08
fix missing edges

kevinkreiser:
it can happen that an edge has the same start and end node as another edge but isnt the same edge. think of the letter thorn (http://en.wikipedia.org/wiki/Thorn_%28letter%29). the loop and the straight edge on the other side of it might be seen as duplicates if you dont account for the edge's index


dgearhart:
:+1:

PR #41, opened 2015-01-08
Incremental progress on hierarchybuilder.

dnesbitt61:
Checking in since changes to use RoadClass in nodes were made.


dgearhart:
:+1:

PR #42, opened 2015-01-08
Fix the node edge index. 

dnesbitt61:
Now successfully placing upward transition edges in the base level.


dgearhart:
when build is good :+1:

PR #43, opened 2015-01-09
Create graph hierarchies

dnesbitt61:
Move the edgeinfo and text logic into GraphTileBuilder so it can be reused between GraphBuilder and HierarchyBuilder. Currently HierarchyBuilder is a separate process from pbfgraphbuilder. It runs pretty fast, but we may want to combine it with pbfgraphbuilder and perhaps add some concurrency to it later.
Creating tiles at each new hierarchy based on the prior hierarchy. Adding upward and downward transition edges.
Still need to add shortcut edges and test within PathAlgorithm, but making progress.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -1,12 +1,6 @@
-#include <boost/functional/hash.hpp>
 #include <future>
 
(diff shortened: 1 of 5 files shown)


dgearhart:
:+1:

PR #44, opened 2015-01-12
Hierarchy Builder updates

dnesbitt61:
Add logic for whether a node can be contracted (shortcut edge formed).
Fixed a bug with edgeindex not being set properly in the new level tiles.


dgearhart:
:+1:

PR #45, opened 2015-01-12
Hierarch Builder fix

dnesbitt61:
Fix the connections so the old tile correctly points to nodes in the new tile.


dgearhart:
:+1:

PR #46, opened 2015-01-13
Use length to differentiate between 2 edges that go between same nodes

dnesbitt61:
Fixes unnamed edge on Lancaster to Pittsburgh route


dgearhart:
:+1:

PR #47, opened 2015-01-14
Add shortcut edges.

dnesbitt61:
This is a working version with some possible enhancements that may lead to more shortcut edges (e.g. test driveability from nodes when considering edge counts).


dgearhart:
I am assuming you wanted to leave the debug in for now :+1:

PR #51, opened 2015-01-16
fix duplicate edgeinfo

kevinkreiser:
it turns out comparing references makes no sense... stupid muscle memory

added a test to prove it was working as it should

also added the ability to load more than one pbf at once

### DIFF
--- a/test/graphtilebuilder.cc
+++ b/test/graphtilebuilder.cc
@@ -1,0 +1,37 @@
+#include "test.h"
+
+#include "mjolnir/graphtilebuilder.h"
+#include <valhalla/baldr/graphid.h>
+#include <valhalla/midgard/pointll.h>
+#include <string>
+#include <vector>
+using namespace std;
+using namespace valhalla::mjolnir;
+
+namespace {
+
+class test_graph_tile_builder : public GraphTileBuilder {
+ public:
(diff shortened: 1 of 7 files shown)


valhalla/mjolnir/graphtilebuilder.h
```diff
@@ -24,8 +24,7 @@
 namespace valhalla {
 namespace mjolnir {
 
-using edge_tuple = std::tuple<const uint32_t,
-    const valhalla::baldr::GraphId&, const valhalla::baldr::GraphId&>;
```
kevinkreiser: this idiocy was the culprit. apologies all

test/graphtilebuilder.cc
```diff
@@ -0,0 +1,58 @@
...
+  //add edge info for node 1 to node 0
+  test.AddEdgeInfo(0, GraphId(0,2,1), GraphId(0,2,0), {{1, 1}, {0, 0}}, {"einzelweg"});
+  if(test.edge_offset_map.size() != 1)
+    throw std::runtime_error("There should still be exactly one of these in here");
+}
+
+}
+
+int main() {
+  test::suite suite("edgeinfobuilder");
```
dgearhart: graph tile buiilder?

dgearhart:
otherwise :+1:

dnesbitt61:
:+1:

PR #52, opened 2015-01-19
Hierarchy Builder fix

dnesbitt61:
Needed to set nextconnectionid to an invalid index once the last connection in the list was used.


dgearhart:
:+1:

PR #53, opened 2015-01-19
Add GraphOptimizer

dnesbitt61:
For now it just sets opposing edge indexes at the end node. Can add logic to it for anything that needs the completed graph (intersection costs, elevation factors, etc.).


gknisely:
ship

PR #55, opened 2015-01-20
Use meters for edge lengths

dnesbitt61:
Update directed edge builder and the graphbuilder/hierarchy builder


gknisely:
:shipit: after circleci good

PR #56, opened 2015-01-20
Gk cycle surface

gknisely:

Have to do paved flag and surface is none is entered.


dgearhart:
:+1:

PR #59, opened 2015-01-21
update logging to use logging framework

kevinkreiser:
please check to make sure all the log levels are appropriate for what they are logging i took my best guess

i did not update the pbf parser as i figure we are going to gut that in the near future anyway for our own solution


dgearhart:
:+1:

gknisely:
ship

PR #60, opened 2015-01-21
Set use for links (either to kRamp or kTurnChannel)

dnesbitt61:
Turn channels are short, never motorway_link or trunk_link, have no exit information (TODO), and each end node connects to a non-link (e.g. neither end node connects only to links - indicating a split or fork).

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -27,9 +27,6 @@
 using namespace valhalla::midgard;
 using namespace valhalla::baldr;
 
-uint32_t rampcount = 0;
-uint32_t turnchannelcount = 0;
-
 namespace valhalla {
 namespace mjolnir {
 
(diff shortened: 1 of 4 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -927,6 +977,9 @@ void GraphBuilder::BuildLocalTiles(const uint8_t level) const {
     }
   }
 
+  std::cout << "Ramp count = " << rampcount << " Turn Channel count = " <<
```
kevinkreiser: use logging

src/mjolnir/graphbuilder.cc
```diff
@@ -737,6 +770,23 @@ void BuildTileSet(std::unordered_map<GraphId, std::vector<uint64_t> >::const_ite
           directededge.set_bridge(w.bridge());
           directededge.set_bikenetwork(w.bike_network());
 
+          // Link - this indicates a ramp or a turn channel. If link is set
+          // test to see if we can infer that the edge is a turn channel
+          directededge.set_link(w.link());
+          if (directededge.link()) {
+            if (directededge.use() != Use::kNone) {
+              std::cout << "Edge is a link but has use = " <<
```
kevinkreiser: use logging

kevinkreiser:
does anyone else feel like graphbuilder is getting to big for its britches? i feel like we should pull out attribution/tag processing to another file and there also seem to be a ton of "helper" functions in there. might they be better suited moving into some of the objects they operate on?

kevinkreiser:
after swapping out cout for LOG_, ship

gknisely:
ship after logging

PR #63, opened 2015-01-22
Shape fixes for HierarchyBuilder

dnesbitt61:
The first edge info added is always forward direction. Needed to reverse the initial shape if the first directed edge was reverse.


kevinkreiser:
looks good to me

gknisely:
:beer:

PR #67, opened 2015-01-23
Fix GetOpposingEdgeIndex

dnesbitt61:
Also rename tileid in HierarchyBuilder (used in 2 locations) to clarify which were base tileids and which were new level tileids.


dgearhart:
:+1:

PR #68, opened 2015-01-24
Update ReclassifyLinks

dnesbitt61:
Trick was to select the best non-link classification at the start node, expand edges until all other nodes with non-link edges are encountered and track the best non-link classification encountered at the end nodes. Then all link edges traversed get assigned the maximum of the best class at the start and the best class at the end (the max means a lesser classification/importance). Had to be careful for cases where the link edges don't actually connect - log these cases.
Also  fix setting not_thru edges - was not adding the proper node to the exand set.


kevinkreiser:
looks good

PR #69, opened 2015-01-26
Add code to check for duplicate edges.

dnesbitt61:
Update ConstructEdges to detect loops (lollipop ways can do this) where startnodeid and endnodeid are
equal. Will no longer add the edge twice when this occurs. This accounted for over 2/3 of the duplicate edges in the PA extract.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -527,8 +527,7 @@ void GraphBuilder::ConstructEdges() {
   uint32_t edgeindex = 0;
   uint64_t startnodeid, nodeid;
   edges_.reserve(edge_count_);
-  for (uint32_t wayindex = 0; wayindex < static_cast<uint32_t>(ways_.size());
-                wayindex++) {
+  for (size_t wayindex = 0; wayindex < ways_.size(); wayindex++) {
     // Get some way attributes needed for the edge
     const auto& way = ways_[wayindex];
 
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -39,8 +39,9 @@ const uint64_t kMaxOSMNodeId = 4000000000;
 // Absurd classification.
 constexpr uint32_t kAbsurdRoadClass = 777777;
 
-uint32_t duplicates = 0;
+// Not thru count and Duplicate way Ids (TODO - put in a stats/issues class)
 uint32_t not_thru_count = 0;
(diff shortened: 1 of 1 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -525,15 +525,16 @@ void GraphBuilder::relation_callback(uint64_t /*osmid*/, const Tags &/*tags*/,
 void GraphBuilder::ConstructEdges() {
   // Iterate through the OSM ways
   uint32_t edgeindex = 0;
-  uint64_t nodeid;
+  uint64_t startnodeid, nodeid;
   edges_.reserve(edge_count_);
-  for (uint32_t wayindex = 0; wayindex < static_cast<uint32_t>(ways_.size()); wayindex++) {
+  for (uint32_t wayindex = 0; wayindex < static_cast<uint32_t>(ways_.size());
```
kevinkreiser: source format be damned! but seriously, why are we using uint32_t here? size_t for all standard containers please

src/mjolnir/graphbuilder.cc
```diff
@@ -858,9 +912,19 @@ void BuildTileSet(std::unordered_map<GraphId, std::vector<uint64_t> >::const_ite
...
+          const Edge& edge = edges[edgeindex];
+          float length = node.latlng().Length(edge.latlngs_);
+          edgelengths.push_back(static_cast<uint32_t>(length + 0.5f));
+        }
+
+        // Look for potential duplicates
+        CheckForDuplicates(osmnodeid, node, edgelengths, nodes, edges, ways);
 
         // Build directed edges
+        uint32_t n = 0;
```
kevinkreiser: if we need n, why not just do a standard for loop instead of for each?

kevinkreiser:
seems like we are doing a lot of statistics in our data creation now. perhaps we could extract all of tha tfunctionality to a statistics class too. might also help in automated reporting of errors in data like you were talking about earlier

src/mjolnir/graphbuilder.cc
dnesbitt61: That was a manual edit to shorten the line. Changing the code to use size_t. Agree with you, with the exception that anything stored in GraphTile should be explicit and use uint64_t  (or 32 or 16 or 8 where it can to save space).

kevinkreiser: oh yeah i figured you would just static cast where it needs to be done. i
guess its 6 and 1/2 dozen of the other or whatever they say?

2015-01-26 8:58 GMT-05:00 David Nesbitt <notifications@github.com>:

PR #71, opened 2015-01-26
make sublogging configs so they can be configured within the same file

kevinkreiser:
this is useful for chef so that we dont ahve to have a seprate config file for every single process we want to run. it also makes it so that pbfgraphbuilder wont bomb if you dont specify logging for it, it will just default to stdout


gknisely:
:shipit:

dgearhart:
:+1:

PR #72, opened 2015-01-26
Hierarchy Builder

dnesbitt61:
Fixed all issues with opposing edge indexes. If you rebuild tiles you should see much better behavior and no errors (at least on the PA extract).


gknisely:
:ship:

PR #73, opened 2015-01-27
Start a DataQuality class. 

dnesbitt61:
Start a class that tracks statistics, issues, and can form quality measures. Very basic so far. Is currently global within GraphBuilder.cc - when we refactor graph building we need to hook it in better.
Also added set_internal to DirectedEdgeBuilder.

### DIFF
--- a/valhalla/mjolnir/dataquality.h
+++ b/valhalla/mjolnir/dataquality.h
@@ -1,0 +1,64 @@
+#ifndef VALHALLA_MJOLNIR_DATAQUALITY_H
+#define VALHALLA_MJOLNIR_DATAQUALITY_H
+
+#include <algorithm>
+#include <unordered_set>
+#include <map>
+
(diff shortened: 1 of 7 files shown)


valhalla/mjolnir/dataquality.h
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * Add statistics for a directed edge.
+   */
+  void AddStats(const baldr::GraphId& tileid, const baldr::DirectedEdge& edge);
+
+  /**
+   * Adds an issue.
+   */
+  void AddIssue(const DataIssueType issuetype, const baldr::GraphId& graphid,
```
kevinkreiser: it would be nice, in my opinion, if this class did the analysis, so we could really move a lot of the logic out of graph builder. this is a good start though!

kevinkreiser:
:clap:

PR #75, opened 2015-01-28
Hierarchy builder improvements

dnesbitt61:
Fix emplace. Add log of tiles updated and created by HierarchyBuilder.
Output list of duplicate ways in separate file (sorted by edge count).

### DIFF
--- a/src/mjolnir/dataquality.cc
+++ b/src/mjolnir/dataquality.cc
@@ -1,4 +1,5 @@
 #include "mjolnir/dataquality.h"
+#include <fstream>
 
 using namespace valhalla::midgard;
 using namespace valhalla::baldr;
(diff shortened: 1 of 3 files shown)


src/mjolnir/dataquality.cc
```diff
@@ -46,16 +47,29 @@ void DataQuality::Log() {
...
+                                    dup.second));
     duplicates += dup.second;
   }
-  LOG_WARN("Duplicate edgecount = " + std::to_string(duplicates));
+  LOG_WARN("Duplicate ways " + std::to_string(duplicateways_.size()) +
+           " duplicate edges = " + std::to_string(duplicates));
+
+  // Sort by edgecount and write to separate file
+  std::ofstream dupfile;
+  std::sort(dups.begin(), dups.end());
```
kevinkreiser: this happens once at the end of building everything right?

kevinkreiser:
looks good

src/mjolnir/dataquality.cc
dnesbitt61: Yea. I wanted to sort the duplicates when the output is logged so I can edit OSM in some priority order

PR #83, opened 2015-01-30
Split parsing from graph building.

dnesbitt61:
Added PBFParser class. Split apart OSMNode and Node.

### DIFF
--- a/src/mjolnir/pbfgraphbuilder.cc
+++ b/src/mjolnir/pbfgraphbuilder.cc
@@ -2,6 +2,7 @@
 #include <vector>
 
 #include "pbfgraphbuilder.h"
+#include "mjolnir/pbfparser.h"
 #include "mjolnir/graphbuilder.h"
 #include "mjolnir/hierarchybuilder.h"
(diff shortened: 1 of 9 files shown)


src/mjolnir/pbfgraphbuilder.cc
```diff
@@ -87,15 +88,49 @@ bool ParseArguments(int argc, char *argv[]) {
...
+  parser.PostProcess();
+  return parser.edge_count();
+}
+
+/**
  * Build local graph from protocol buffer input.
  */
 void BuildLocalGraphFromPBF(const boost::property_tree::ptree& pt,
                const std::vector<std::string>& input_files) {
+  // Stores all the ways that are part of the road network
```
kevinkreiser: can we wrap all of thse in a struct defined within PBFParser, and have  the parser return a const ref to the struct which we can then pass along to graphbuilder::build?

kevinkreiser:
looks good other than comment above

dgearhart:
:+1:

gknisely:
:ship:

PR #84, opened 2015-01-30
OSMData

dnesbitt61:
USe OSMData (simple container structure) to pass data between PBFParser and GraphBuilder.


kevinkreiser:
love it :rocket:

PR #89, opened 2015-02-02
Use std::pair for internal LL storage

dnesbitt61:
Use std::pair of floats for lat,lng. Add a method to get shape as a vector of PointLL so we can call Length. Update NodeInfoBuilder to remove PointLL and set LL as std::pair.


dgearhart:
:+1:

PR #93, opened 2015-02-03
Add GraphId to GraphTileHeaderBuilder.

dnesbitt61:
Also change to edgeinfo_offset in directed edge builder and places where it is set.


dgearhart:
:+1:

PR #97, opened 2015-02-04
Init relations gk

gknisely:

Init logic for restrictions.  Have to process day and time logic yet.


dgearhart:
:+1:

PR #105, opened 2015-02-06
Exit signs were being added to shortcut edges - do not want that.

dnesbitt61:
Also fix the index in the loop inside ConnectBaseLevelToNewLevel
(was using n which was being used for number of added transition edges).

### DIFF
--- a/src/mjolnir/hierarchybuilder.cc
+++ b/src/mjolnir/hierarchybuilder.cc
@@ -163,12 +163,20 @@ bool HierarchyBuilder::CanContract(const GraphTile* tile, const NodeInfo* nodein
     }
   }
 
-  // Mark the 2 edges entering the node (opposing) and 2 edges exiting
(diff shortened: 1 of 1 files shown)


gknisely:
:+1:

src/mjolnir/hierarchybuilder.cc
```diff
@@ -647,14 +661,18 @@ void HierarchyBuilder::AddConnectionsToBaseTile(
 
     // Add existing directed edges
     uint32_t idx = node.edge_index();
-    for (uint32_t n = 0; n < node.edge_count(); n++) {
+    for (uint32_t j = 0; j < node.edge_count(); j++) {
```
dgearhart: is j used?

dnesbitt61: only as the loop index to control the number of iterations

PR #107, opened 2015-02-07
Switched to use capitalized directional i.e. US 30 W >> US 30 West

dgearhart:
Also, updated logic when adding exit info to directed edge

@dnesbitt61 @gknisely @kevinkreiser


gknisely:
you changed your mind again!  :+1:

PR #111, opened 2015-02-11
Data tests gk

gknisely:
Not the end of the data tests, but the beginning.
Also, fixed access issue for nodes.

@kevinkreiser @dgearhart @dnesbitt61


dgearhart:
:+1:

PR #112, opened 2015-02-12
Updates to lessen memory use of Node and Edge.

dnesbitt61:
Also add shrink_to_fit for a variety of vectors. Edge now stores indexes into a single lat,lng list and Node no longer stores a GraphId as the Nodes are created and tiled within ConstructEdges, and the tile location servers as the GraphId. Node no longer derives from OSMNode - this leads to some size savings. Replaced vector with list for the Edges from each Node - that lowered memory size a little as well.


dgearhart:
:+1:

gknisely:
:ship:

src/mjolnir/graphbuilder.cc
```diff
@@ -261,118 +280,122 @@ uint32_t GraphBuilder::GetBestNonLinkClass(const Node& node) const {
...
-      // Expand from all link edges
-      for (auto startedgeindex : node.second.edges()) {
-        // Get the edge information. Skip non-link edges
-        const Edge& startedge = edges_[startedgeindex];
-        if (!startedge.attributes_.fields.link) {
-          continue;
-        }
+  std::unordered_set<GraphId> visitedset;  // Set of visited nodes
+  std::unordered_set<GraphId> expandset;   // Set of nodes to expand
+  std::vector<uint32_t> linkedgeindexes;   // Edge indexes to reclassify
```
kevinkreiser: make this a list and it will be faster and use less memory

src/mjolnir/graphbuilder.cc
```diff
@@ -37,222 +36,242 @@ constexpr uint32_t kMaxNoThruTries = 256;
...
-  LOG_INFO("Sizeof OSM node refs: " + std::to_string(osmdata.noderefs.size()));
-  delete_noderefs(osmdata);
-  LOG_INFO("After delete: Sizeof OSM node refs: " + std::to_string(osmdata.noderefs.size()));
+  // Shrink latlng and edge vectors
+  latlngs_.shrink_to_fit();
+  edges_.shrink_to_fit();
+
+  // Iterate through the tilednodes and shrink vectors
+  for (auto& tile : tilednodes_) {
+    tile.second.shrink_to_fit();
```
kevinkreiser: i wonder if the tilednodes even need to be a vector. dont we just loop over them in the thread anyway? they could just be a list and we'll never be overcommitted on memory or have to pay for realloc

PR #114, opened 2015-02-13
Add plumbing for adding simple turn restrictions

dnesbitt61:
Lots of changes to identify simple TRs and make it so they can be added. Next need to actually add them and update the hierarchy builder to properly update them and handle them during shortcut creation.


gknisely:
:ship:

PR #115, opened 2015-02-13
Add simple turn restrictions (part 2)

dnesbitt61:
This PR actually adds them to the tile. HierarchyBuilder updates edge indexes for the TRs during hierarchy building. All that is left to do is to update TRs on the arterial and highway levels and to figure out how they impact shortcut edges (not that these are trivial tasks!)

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -278,7 +278,9 @@ void GraphBuilder::UpdateRestrictions(OSMData& osmdata) {
   for (auto& rst : osmdata.restrictions) {
     const auto nd = nodes_.find(rst.second.via());
     if (nd == nodes_.end()) {
-      LOG_INFO("Restriction Via node on a non-graph node");
+      // TODO - log these as data issues??
+      LOG_ERROR("Restriction Via node on a non-graph node: from wayid = "
+           + std::to_string(rst.first));
     } else {
       rst.second.set_via(nd->second);
     }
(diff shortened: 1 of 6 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -660,6 +662,96 @@ void CheckForDuplicates(const GraphId& nodeid, const Node& node,
...
+      } else {
+        trs.push_back(r->second);
+      }
+    }
+  }
+  if (trs.empty()) {
+    return false;
+  }
+
+  // TODO - cannot mix types (no and only!)
```
kevinkreiser: looks like you took care of this todo right below?

src/mjolnir/graphoptimizer.cc
```diff
@@ -24,6 +24,38 @@ GraphOptimizer::GraphOptimizer(const boost::property_tree::ptree& pt)
...
+          if (directededge->exitsign()) {
+            if (tile.GetSigns(idx).size() == 0) {
+              LOG_ERROR("Directed edge marked as having signs but none found");
+            }
+          }
+        }
+      }
+    }
+  }
+  LOG_INFO("DONE");
```
kevinkreiser: ive been thinking we need a start and done message for the whole tile building process, since the log hits the same file (especially on amazon) its nice to see start and done messages. of course especially if they have something unique to them, so maybe here we could say "validation done"?

src/mjolnir/graphbuilder.cc
dnesbitt61: yea, fixed the comment and will update in next PR.

kevinkreiser:
:ship:

PR #116, opened 2015-02-13
Admin logic gk

gknisely:

Init logic for admins.  Next, need to save to sqlite


valhalla/mjolnir/osmadmin.h
```diff
@@ -0,0 +1,116 @@
...
+
+  // Count of members.
+  uint16_t memberid_count_;
+
+  // Name of country or state/prov
+  std::string name_;
+
+  // Admin level.  2 = country; 4 = state.
+  uint8_t admin_level_;
+};
```
dnesbitt61: Might get smaller size/better data packing if move admin_level above string

dnesbitt61:
Looks fine. Do we get all the admin information we need from OSM or are there different/better sources?

valhalla/mjolnir/osmadmin.h
gknisely: Ok will do.

gknisely:
Should have all the info we need minus timezone info.

dgearhart:
:+1:

PR #119, opened 2015-02-17
Changes to use vector rather than unordered_map for OSM nodes

dnesbitt61:
Allows all of North America to be loaded into osmdata and still be below 10GB. 
Also hardcoded to 1 thread for now - threading seems to incur swap more readily.
The node tests are commented out...not sure what the behavior will be if the node is not found.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -140,7 +140,7 @@ Node& GraphBuilder::GetNode(const GraphId& graphid) {
 }
 
 // Construct edges in the graph and assign nodes to tiles.
-void GraphBuilder::ConstructEdges(OSMData& osmdata, const float tilesize) {
+void GraphBuilder::ConstructEdges(const OSMData& osmdata, const float tilesize) {
   // Reserve size for the Node map
   nodes_.reserve(osmdata.intersection_count);
 
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -50,7 +50,8 @@ GraphBuilder::GraphBuilder(const boost::property_tree::ptree& pt)
 
 // Delete the OSM node map and extended node information maps.
 void delete_osmnode_map(OSMData& osmdata) {
-  OSMNodeMap().swap(osmdata.nodes);
+//  OSMNodeMap().swap(osmdata.nodes);
+  std::vector<OSMNode>().swap(osmdata.nodes);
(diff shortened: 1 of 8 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -1107,7 +1121,7 @@ std::vector<SignInfo> GraphBuilder::CreateExitSignInfoList(
 // Build tiles for the local graph hierarchy
 void GraphBuilder::BuildLocalTiles(const uint8_t level, const OSMData& osmdata) const {
   // A place to hold worker threads and their results, be they exceptions or otherwise
-  std::vector<std::shared_ptr<std::thread> > threads(threads_);
+  std::vector<std::shared_ptr<std::thread> > threads(1); //threads_);
```
kevinkreiser: i think you meant to change this back. if you want to run mjolnir with only 1 thread set it that way in the config

gknisely:
:ship: after u fix what kevin noted for the config.

valhalla/mjolnir/osmdata.h
```diff
@@ -64,6 +66,17 @@ struct OSMData {
   // Names
   UniqueNames name_offset_map;
 
+  OSMNode GetNode(const uint64_t osmid) {
```
kevinkreiser: should make this method const

src/mjolnir/graphbuilder.cc
```diff
@@ -138,7 +140,7 @@ Node& GraphBuilder::GetNode(const GraphId& graphid) {
 }
 
 // Construct edges in the graph and assign nodes to tiles.
-void GraphBuilder::ConstructEdges(const OSMData& osmdata, const float tilesize) {
```
kevinkreiser: should pass this by const ref after making GetNode const

valhalla/mjolnir/osmdata.h
```diff
@@ -64,6 +66,17 @@ struct OSMData {
...
+    OSMNode test(osmid, 0.0f, 0.0f);
+    auto it = std::equal_range(nodes.begin(), nodes.end(), test);
+    return *(it.first);
+  }
+
+  // Sort the OSM nodes
+  void SortNodes() {
```
kevinkreiser: i think we should remove this method, and then in pbfgraphparser.load, do it directly there. that way anyone who gets an osmdata object won't have to know to call this method.

dnesbitt61: yea, was thinking that too...

PR #120, opened 2015-02-18
Added processing of destination:street and destination:street:to and Fixed bug with OSMWay::destination_ref_index

dgearhart:
@dnesbitt61 @kevinkreiser @gknisely


gknisely:
:ship:

PR #121, opened 2015-02-18
Rework simple turn restrictions

dnesbitt61:
Part of the directededge rather than a separate lookup in the tile data.


gknisely:
:ship: after circle ci good

PR #125, opened 2015-02-20
Fix for loops in parser. Update turn channel and restriction logic.

dnesbitt61:
Reserver and trim methods wre inside the loop (over each input file) - moved these out and also updated timing and logging to be outside the loop.
Also get node LL only on first edge (was unnecessarily getting it on ever edge) in the graph builder.
Update to check if restriction mask exceeds the max. and log error and constrain it if so.


kevinkreiser:
looks ok to me but to be honest i dont understand some of the stuff going on with turn channel and restriction masks just yet

dgearhart:
:+1:

PR #126, opened 2015-02-20
Updates for ref bollard

gknisely:
@dgearhart 's ref updates.  removable bollards now saved as bollards and not gates.

@dnesbitt61 @kevinkreiser


dgearhart:
:+1:  thanks

PR #132, opened 2015-02-24
Add set methods for new data members. 

dnesbitt61:
Update comments and method order to match baldr/directededge.h


dgearhart:
:+1:

PR #133, opened 2015-02-24
Sets Up the Parsing for Multiple Threads

kevinkreiser:
this just does the rearranging needed to actually add multiple threads for parsing. the next pr will actually add this. after that i'll go back and add kyotocabinet for the osmnodes storage (which to be economical kind of requires having threaded parsing)


dgearhart:
:+1:

gknisely:
:ship:

PR #136, opened 2015-02-25
Traffic signals

gknisely:


src/mjolnir/graphbuilder.cc
```diff
@@ -896,9 +899,21 @@ void BuildTileSet(
...
+          if (!forward && node.traffic_signal())
+            has_signal = true;
+          // traffic signal exists at a non-intersection node
+          else if (edge.attributes.traffic_signal) {
+            if (forward && edge.attributes.forward_signal)
+              has_signal = true;
+            else if (!forward && !edge.attributes.forward_signal)
+              has_signal = false;
+          }
+
```
dnesbitt61: can simplify since you set has_signal to false no need to set it again

PR #139, opened 2015-02-27
Add GraphEnhancer- populate node density

dnesbitt61:
Computes node density - creates 0-15 values by compressing kilometers fo road per swuare km. May want to play around with the mapping of density into the relative value and we may actually want  2 values, one a more local density (say within 2k) vs a wider density (10km). At any rate provides the framework for further enhancements to the local level graph.


kevinkreiser:
:shipit: can you add this to the test in tyr as well where it builds test tiles?

kevinkreiser:
https://github.com/valhalla/tyr/blob/master/test/handlers.cc#L60 appx here or so?

PR #140, opened 2015-03-02
reduce memory use

kevinkreiser:
this sends `OSMNode`s (which contains the shape) to file, `OSMWay`s to file and graph `Edge`s to file. for PA the time went up from 1min 8 sec to 2 min 1 sec. The files it created for PA seemed to take up about 4x the original pbf file's size, which shouldnt matter for even the planet given that its 27 gigs (so we'd need like 150gb of disk space to do it). The only thing of any real weight in memory at this point are the strings constituting the names of various stuff and the `Node`s in the graph (since they arent fixed size we can't move them to file very easily).

this build is failing because it exploits the fact that osm extracts are sorted for each set of items (ways nodes relations) by the given items id. the extracts we use for the parsing tests aren't sorted and the code checks this and throws an exception. however they were created with xml extracts which @gknisely is sending me. i'll sort those and turn them into pbf again and we should be good to go!

and one more thing, enjoy the festive christmas diff!


dnesbitt61:
I'm good with this if it converts all of Americas! And can then do the office to PNC Park (Pittsburgh) route - 380.4 km...

gknisely:
holy :christmas_tree:

dgearhart:
@gknisely  rockefeller :christmas_tree:

PR #145, opened 2015-03-04
Stats per thread. Temporary fix to ReclassifyLinks

dnesbitt61:
Fix solves the I-83/I-695 E transition. 
Also inferring cul-de-sacs.


kevinkreiser:
other than the future.get() happening twice looks great!

PR #146, opened 2015-03-04
thread the graphenhancer

kevinkreiser:
made the enhance function static. actually id like to make all of the main pieces of mjolnir static as it doesnt make sense to use the objects anyway after you've used them once. so far parsing and enhancing are static. we should make graphbuilder and hierarchy builder static too at some point IMHO

enhancement stage with 8 threads is down to 45 seconds compared to 170 seconds single threaded


dgearhart:
sweet :+1:

gknisely:
:ship:

PR #147, opened 2015-03-04
fixed SA crash.

gknisely:

### DIFF
--- a/conf/edges.lua
+++ b/conf/edges.lua
@@ -514,9 +514,9 @@ function rels_proc (kv, nokeys)
        local restrict = restriction[kv["restriction"]]
        if restrict ~= nil then
          kv["restriction"] = restrict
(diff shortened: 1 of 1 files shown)


kevinkreiser:
any chance for a description on this? what will happen when we encounter a restriction that looks like:

"Mo-Fr 07:00-18:10"

gknisely:
It will pass now.  The above was for a bus route and the data is entered incorrectly.

gknisely:
Both fields should have data.  day_on and day_off

kevinkreiser:
so this data is poorly tagged and will be ignored?

gknisely:
yes sir

kevinkreiser:
:sheep: it

dgearhart:
:+1:

PR #148, opened 2015-03-05
Admin logic gk

gknisely:

So after fighting with the osm data (inner and outer polygon good times),multipolygons , order of ways in the relation, spatialite (mixing polygons and multipolygons in a geometry type column is not allowed), etc...I have decided to port some of the osm2pgsql geos code to pbfadminbuilder.  We can remove this at a later date if we can solve the order issues.  

I was able to process the planet with vectors until the mega Kevin merge; however, switching to maps is sucking up too much memory.  After this merge, I will most likely go back to vectors. 

Next up is  make sure I can process the planet again and same the info to the tiles.
This picture is the first pass....I allowed states for all countries...now I do not.
screenshot from 2015-03-05 14 46 40

Northeast:
screenshot from 2015-03-05 14 58 13


dgearhart:
CI not happy

valhalla/mjolnir/osmdata.h
```diff
@@ -71,6 +75,15 @@ struct OSMData {
   // Names
   UniqueNames name_offset_map;
 
+  // Map used in admins to store the shape of the nodes.
```
kevinkreiser: i keep wondering if stuff like this belongs here. osmdata seems to be meant to be the interface between graphparser and graphbuilder. perhaps these would be better suited to live in the admin callback or perhaps derive another struct from osmdata that only the adminparser knows about and uses? unless we are thinking of doing these at the same time, in which case this would make sense and admin parsing should merge with graph parsing? i dont know haha

kevinkreiser:
too much memory with maps? that seems kind of nuts. the osm coastlines file, was only 300 megs compressed. it didnt have all of the state level admins but still, do you think that uncompressed it would be > 500x the size of coastlines file?

valhalla/mjolnir/osmdata.h
gknisely: I agree and I was thinking the same thing.  I use the uses to not add nodes and ways that are not needed.  But the other stuff is not needed.  I would like to get circle up and this merged.  Then I can work on that.

gknisely:
Ok...finally circle good with the help of @kevinkreiser  Merging and will look at memory use.

PR #149, opened 2015-03-06
when asking for tiles that may or may not be there check first before you use them

kevinkreiser:
NA is enhancing now. If it completes we are good as it always segfaulted on nullptr tile for me before


dgearhart:
:+1:

PR #151, opened 2015-03-06
Set methods to reflect changes to NodeInfo and DirectedEdge

dnesbitt61:
Added some tests due to bit field manipulation (and added a convenience method to the builders to overwrite a set of bits).


dgearhart:
:shipit:  when CI is happy

PR #152, opened 2015-03-08
Fix opp_local_idx for shortcut edges.

dnesbitt61:
It needs to be the opp_local_idx of the last directed edge in the shortcut chain.


dgearhart:
:shipit:

gknisely:
:ship:

PR #153, opened 2015-03-09
only cleanup pbf once after finishing all the parsing

kevinkreiser:
we still want to cleanup becuse we want as much mem as we can get for the actual tile building at least in th ecase of graphtiles, mayb enot so much for admins but whatever


dgearhart:
:shipit:

PR #157, opened 2015-03-10
Add edge_to_left and edge_to_right

dnesbitt61:
Verified with point and click using Tyr (with a log statement in trip path builder).
Also updated some comments and TODOs.


dgearhart:
:shipit:

PR #161, opened 2015-03-11
Updates to shortcut and superseded to make them a mask rather than bool

dnesbitt61:
Fix in hierarchy builder. Fix GraphOptimizer to only check if a shortcut (not its mask-but bool value saying if shortcut exists)


dgearhart:
:shipit:

gknisely:
:ship:

PR #162, opened 2015-03-12
send graph nodes to file

kevinkreiser:
this lets us safe s-tons of memory. a this point the only thing in memory of any weight is the street name map. this does use memory maps for everything though so the os will pretty much let you access the file as if it were memory up to the point of running out of memory but just not, which is what we want.

when writing this we realized that we werent adding a second directed edge for loops. i have put code in place to do this but i have it commented out because its not working quite right.

on the pa import the whole thing still takes only 2 minutes which is great because it basically doesnt increase the import time at all.

I guess i should say how it works!

When constructing edges we put the start and end graph node of a given graph edge to file (so the nodes at intersections are duplicated). Each node stores a reference to the edge from which it was born (and whether it was the beginning of the edge or end). We then sort the graph nodes first by graphid (what tile they are in) and then by osmid so we have all of the duplicate nodes next to eachother, like this (these are osmids of nodes):

3,3,3,3,7,7,7,9,25,25

Then when we ware writing tiles we simply scan seqentially through the file, node 3 has 4 edges connected to it, node 7 has 3 edges, node 9 is a dead end, node 25 might be a gate. For any given import we now have 4 files we memory map:

ways.bin - with just wayid and attribution per way
way_nodes.bin - with each node and its id and attribution as shape points of the way in the order they occur in the way (duplicates where ways intersect)
edges.bin - the ways broken up into graph edges where the ways intersect
nodes.bin - the start and end node of each graph edge, does encapsulate the way_node and references to the graph edge in the edges file

For the PA import (pbf is only 106M) we end up with:
kkreiser@klaptop:~/sandbox/valhalla/mjolnir$ ls -hal *.bin
-rw-r--r-- 1 kkreiser kkreiser  27M Mar 12 09:14 edges.bin
-rw-r--r-- 1 kkreiser kkreiser  76M Mar 12 09:14 nodes.bin
-rw-r--r-- 1 kkreiser kkreiser 290M Mar 12 09:14 way_node_ref.bin
-rw-r--r-- 1 kkreiser kkreiser  47M Mar 12 09:14 ways.bin

So we slightly more than quadrupale the original import size.. This would mean that for a planet import we'll need about 28G * 4.15 = 116G

### DIFF
--- a/src/mjolnir/directededgebuilder.cc
+++ b/src/mjolnir/directededgebuilder.cc
@@ -2,6 +2,8 @@
 #include <valhalla/baldr/nodeinfo.h>
 #include <valhalla/midgard/logging.h>
 
+#include <algorithm>
+
 namespace {
 
 /**
@@ -494,6 +496,13 @@ void DirectedEdgeBuilder::set_edge_to_right(const uint32_t localidx,
     stopimpact_.edge_to_right = OverwriteBits(stopimpact_.edge_to_right,
                                 right, localidx, 1);
   }
+}
+
+DirectedEdgeBuilder DirectedEdgeBuilder::flipped() const {
+  auto other = *this;
+  std::swap(other.forwardaccess_, other.reverseaccess_);
+  other.attributes_.forward = !other.attributes_.forward;
+  return other;
 }
 
 }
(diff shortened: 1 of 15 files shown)


dnesbitt61:
:shipit:

dgearhart:
:shipit:

gknisely:
:christmas_tree:

PR #163, opened 2015-03-12
further remove the object oriented-ness from the various stages of tile ...

kevinkreiser:
...building so as to emphasize the fact that they are once only operations with state that only lives on disk. this also means that we dont keep memory around for their life times because they have no lifetimes

this looks like  a christmas tree diff but its basically just moving the functions around. @dnesbitt61 if this is going to screw you up i can wait to merge it until you are done on your branch. or i could just merge it into your branch and then we can get it when you pr next.


dnesbitt61:
:shipit:

PR #166, opened 2015-03-13
Move not_thru logic into GraphEnhancer. 

dnesbitt61:
Update shortcut logic to disable incoming driveability check - fix Breezewood issue


kevinkreiser:
:rocket:

PR #167, opened 2015-03-13
route tiles for the whole the planet!!!!

kevinkreiser:
This enables cache size control during the later stages of tile refinement in which it is presumed that we are running out of memory because we currently have an infinite cache limit. This sets a limit by default (1 gig) and checks periodically if we are over that limit and just purges the whole cache. Later on we will need to be smarter about this, but that will happen when we put the cache api behind a service as well anyway so this should be ok for now.


dnesbitt61:
:ship:

PR #170, opened 2015-03-16
get the right graph  node count

kevinkreiser:
this can be removed later when we do full stats tracking at the tile level


dnesbitt61:
:ship:

PR #173, opened 2015-03-17
Fix for ReclassifyLinks

dnesbitt61:
Was including an emergency access only service road in the classification process. This prevented taking I-295 around Richmond,VA


dgearhart:
:shipit:

PR #174, opened 2015-03-18
Fix for when 2 incoming ramps join a highway. 

dnesbitt61:
Was not setting the classification. Fixes I-81 onto I-84 case near Wilkes Barre, PA


dgearhart:
:shipit:

PR #177, opened 2015-03-20
Admins

gknisely:

First pass at admins.  Will skip admins if db does not exist.


gknisely:
Depends on baldr

dgearhart:
:shipit:

PR #179, opened 2015-03-30
Add ctry crossing flag and update edge sorting

dnesbitt61:
Edges from a node are sorted by: 1) driveability from the node, 2) importance, 3) presence of names.
Ctry crossing flag added but not set yet (placeholder in graphenhancer for when admin logic is more complete).


gknisely:
:rocket:

PR #180, opened 2015-03-31
Update density method, adjust some urban/rural speeds

dnesbitt61:
Add MPH comments so I don't get confused again! No longer have separate local density - just use density for rural/urban checks - that way we can debug more easily since we can check what density is assigned per edge.


gknisely:
:+1:

PR #183, opened 2015-04-06
Admin updates

gknisely:
@kevinkreiser @dgearhart @dnesbitt61 

admin updates:  Includes state and country names(local) and iso codes.


dgearhart:
:shipit:

PR #188, opened 2015-04-07
Fix operator precedence.

fsaric:
Operator & has lower precedence than !=.


dnesbitt61:
Thanks - good eye. Plus it is always good to be explicit so the added parenthesis do that.

PR #194, opened 2015-04-13
Add relative tile density.

dnesbitt61:
Now where to put it...question at different levels - how do we normalize (since roads drop out at different levels).


dgearhart:
:shipit:

PR #197, opened 2015-04-16
Add some density stats (average and max)

dnesbitt61:
so we can figure out how to assign relative density measures.


dgearhart:
:shipit:

gknisely:
:ship:

PR #200, opened 2015-04-20
tile backup and clean up scripts for chef.

gknisely:


scripts/backup_tiles.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+TIME=`date +%Y_%m_%d_%H%M%S`
+rsync -avPc $1 $1.$TIME
```
kevinkreiser: i might not be understanding the big picture but... why copy them? why not just move them? or better yet when cutting new tiles why not just cut new ones to a tmp dir and only move them in once it succeeds?

PR #211, opened 2015-04-27
Move link Use and speed updates into GraphEnhancer

dnesbitt61:
stats for PA match. Logic is a bit simpler and can now check for presence of exit signs as well.


dgearhart:
:shipit:

PR #212, opened 2015-04-28
Update ReclassifyLinks

dnesbitt61:
Update ConstructEdge to flag short link edges. Update ReclassifyLinks to use the shortlink flag to expand at nodes with non-link edges only if a short link is present (likely to be internal to an intersection)


dgearhart:
:shipit:

PR #213, opened 2015-04-29
Intermediate work on transit. 

dnesbitt61:
Stop queries are in place, framework started for adding to tiles

### DIFF
--- a/src/mjolnir/transitbuilder.cc
+++ b/src/mjolnir/transitbuilder.cc
@@ -1,11 +1,12 @@
 #include "mjolnir/transitbuilder.h"
-#include "mjolnir/idtable.h"
 #include "mjolnir/graphtilebuilder.h"
 
 #include <list>
 #include <future>
 #include <thread>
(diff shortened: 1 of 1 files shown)


src/mjolnir/transitbuilder.cc
```diff
@@ -51,49 +102,292 @@ void GetStops(sqlite3 *db_handle, const AABB2& aabb) {
...
   }
   if (stmt) {
+    sqlite3_finalize(stmt);
+    stmt = 0;
+  }
+  return stops;
+}
+
+// Lock on queue access since we are using it in different threads. No need
+// to lock graphreader since no threads are writing tiles yet.
```
kevinkreiser: if two threads try to load get at an uncached tile through graphreader at the same time you could crash (because both will be trying to write to the map it uses). so even though you are only reading from disk, you are still writing to memory from separate threads. only way you can be sure its call good is if its a const ref.

PR #214, opened 2015-04-29
Fix bug where edges with exitsigns were not marked.

dnesbitt61:
Check if a link, not Use = ramp. Ramp/turn channel use assignment is now in GraphEnhancer


dgearhart:
:shipit:

dgearhart:
:+1:

PR #216, opened 2015-04-30
Additions to create nodes and directed edges in the graph (for transit).

dnesbitt61:
Still need to add to the tile and build the new tile, but getting closer.


dgearhart:
:shipit:

PR #219, opened 2015-05-04
Transit additions. Method to add names to textlist

dnesbitt61:
Adding in transit information (except nodes and directed edges).
Add a flag to GraphTileBuilder constructor that triggers logic to read in tile data such that edges and text can be added to.
Add method (AddNames) that allows names to be added to text list. Use this for transit names and also signs, edgeinfo, and admin were ported to use it.

### DIFF
--- a/src/mjolnir/graphenhancer.cc
+++ b/src/mjolnir/graphenhancer.cc
@@ -362,8 +362,9 @@ bool IsIntersectionInternal(GraphReader& reader, std::mutex& lock,
                             DirectedEdgeBuilder& directededge,
                             const uint32_t idx) {
   // Internal intersection edges must be short and cannot be a roundabout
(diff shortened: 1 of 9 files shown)


gknisely:
:train:

src/mjolnir/graphenhancer.cc
```diff
@@ -362,8 +362,9 @@ bool IsIntersectionInternal(GraphReader& reader, std::mutex& lock,
                             DirectedEdgeBuilder& directededge,
                             const uint32_t idx) {
   // Internal intersection edges must be short and cannot be a roundabout
```
dgearhart: update comment based on new logic

dgearhart:
otherwise :shipit:

PR #221, opened 2015-05-05
Add transit nodes and edges

dnesbitt61:
Add nodes for stops.
Add edges for parent/child  connections. These must have exact pairs of opposing edges.
Add edges for unique route/stop pair departures. These currently do NOT have opposing edges.
TODO - shape for edges, names (route names?), and connections to road network.


dgearhart:
:shipit:

PR #222, opened 2015-05-07
added seconds_from_midnight and days_from_pivot_date

gknisely:


gknisely:
depends on baldr

dgearhart:
:shipit:

PR #223, opened 2015-05-08
Added IsPencilPointUturn method so we could identify pencil point u-turns and increase the stop impact so we can avoid if possible.

dgearhart:


dnesbitt61:
:shipit:

PR #235, opened 2015-05-15
added logic to process unsigned_ref.

gknisely:


dgearhart:
nice, thanks :shipit:

PR #236, opened 2015-05-18
Add empty tiles (header only) when way intersects tile with no nodes

dnesbitt61:
This is used so we can generate a valid connectivity map.
Have to skip tiles with 0 nodes when forming density or hierarchy.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -911,7 +911,7 @@ void BuildLocalTiles(const unsigned int thread_count, const OSMData& osmdata,
     threads[i].reset(
       new std::thread(BuildTileSet,  std::cref(ways_file), std::cref(way_nodes_file),
                       std::cref(nodes_file), std::cref(edges_file), std::cref(tile_hierarchy),
-                      std::cref(osmdata), std::ref(tiles), tile_start, tile_end,
+                      std::cref(osmdata), std::cref(tiles), tile_start, tile_end,
                       std::ref(results[i]))
     );
   }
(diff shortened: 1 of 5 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -843,7 +911,8 @@ void BuildLocalTiles(const unsigned int thread_count, const OSMData& osmdata,
     threads[i].reset(
       new std::thread(BuildTileSet,  std::cref(ways_file), std::cref(way_nodes_file),
                       std::cref(nodes_file), std::cref(edges_file), std::cref(tile_hierarchy),
-                      std::cref(osmdata), tile_start, tile_end, std::ref(results[i]))
+                      std::cref(osmdata), std::ref(tiles), tile_start, tile_end,
```
kevinkreiser: shouldn't that be cref?

src/mjolnir/graphbuilder.cc
```diff
@@ -867,6 +936,19 @@ void BuildLocalTiles(const unsigned int thread_count, const OSMData& osmdata,
       //TODO: throw further up the chain?
     }
   }
+
+  // Add "empty" tiles for any intersected tiles
+  for (auto& empty_tile : stats.intersected_tiles) {
+    if (GraphReader::DoesTileExist(tile_hierarchy, empty_tile)) {
```
kevinkreiser: instead of passing in the full map of tiles that are going to be written to each thread and  having them only add blank tiles to the interesected_tiles list when its not in the main one, i think it woudl be cleaner to just put all of them into that and then in this if statement check if they are in the main map of tiles or not. less passing stuff around. easier to read, less complicated CheckForIntersectingTiles function

PR #241, opened 2015-05-21
Connect transit stops to OSM edges

dnesbitt61:
Data conversion with no errors for CA with BART transit DB present. Verified that regular routes work - next step is to validate and debug transit routing!

NOTE: stub for a method to reclassify edges that connect to ferries is in place but is incomplete and not actually used yet.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -61,7 +61,8 @@ struct Edge {
     uint32_t shortlink        : 1;   // true if this is a link edge and is
                                      //   short enough it may be internal to
                                      //   an intersection
-    uint32_t spare            : 3;
+    uint32_t ferry            : 1;
+    uint32_t spare            : 2;
   };
   EdgeAttributes attributes;
 
@@ -92,6 +93,7 @@ struct Edge {
       e.attributes.driveablereverse = way.auto_backward();
     }
     e.attributes.link = way.link();
+    e.attributes.ferry = way.ferry() || way.rail();
     e.attributes.reclass_link = false;
     e.attributes.has_names = (way.name_index_ != 0
                            || way.name_en_index_ != 0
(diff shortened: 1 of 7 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -939,9 +1055,7 @@ void BuildLocalTiles(const unsigned int thread_count, const OSMData& osmdata,
 
   // Add "empty" tiles for any intersected tiles
   for (auto& empty_tile : stats.intersected_tiles) {
-    if (GraphReader::DoesTileExist(tile_hierarchy, empty_tile)) {
-      LOG_ERROR("Writing an empty tile over a valid tile!");
-    } else {
+    if (!GraphReader::DoesTileExist(tile_hierarchy, empty_tile)) {
```
kevinkreiser: instead of checking the disk here you could just do `tiles.find(empty_tile) == tiles.cend()`

should be faster but honestly with so few tiles in total it wont make much difference

src/mjolnir/transitbuilder.cc
```diff
@@ -607,21 +747,70 @@ void AddToGraph(GraphTileBuilder& tilebuilder,
...
+                        tile->header()->graphid().level(), i);
+          endnode = directededge->endnode();
+          mindist = std::get<1>(closest);
+        }
+      }
+    }
+  }
+
+  // TODO - more complete shape
+  std::vector<PointLL> shape;
```
kevinkreiser: this can just be `std::vector<PointLL> shape{tile->node(startnode)->latlng(), stop.ll}`

src/mjolnir/transitbuilder.cc
```diff
@@ -697,12 +886,35 @@ void build(const boost::property_tree::ptree& pt,
...
+    std::vector<OSMConnectionEdge> connection_edges;
+    for (auto& stop : stops) {
+      if (stop.graphid.Tile_Base() == tile_id) {
+        // Add connections to the OSM network
+        if (stop.parent == 0) {
+          AddOSMConnection(stop, tile, connection_edges);
+        }
+      }
+    }
+LOG_INFO("Connection Edges: size= " + std::to_string(connection_edges.size()));
```
kevinkreiser: why no indent?

kevinkreiser:
other than comments ship

gknisely:
:train: after fix

src/mjolnir/transitbuilder.cc
dnesbitt61: that is my cue to remove later!

PR #242, opened 2015-05-22
Use lineIds for departures

dnesbitt61:
Other assorted changes that now allow us to find departures properly.

### DIFF
--- a/src/mjolnir/transitbuilder.cc
+++ b/src/mjolnir/transitbuilder.cc
@@ -265,7 +265,7 @@ void assign_graphids(const boost::property_tree::ptree& pt,
     // the map/list
     std::vector<Stop> stops = GetStops(db_handle, tiles.TileBounds(id));
     if (stops.size() > 0) {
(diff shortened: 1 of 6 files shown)


src/mjolnir/transitbuilder.cc
```diff
@@ -734,6 +735,13 @@ LOG_INFO("Add conn from stop to OSM: ei offset = " + std::to_string(edge_info_of
       directededge.set_localedgeidx(directededges.size());
       directededge.set_pedestrianaccess(true, true);
       directededge.set_pedestrianaccess(false, true);
+      directededge.set_lineid(transitedge.lineid);
+
+      LOG_INFO("Add directededge - lineId = " + std::to_string(transitedge.lineid) +
```
kevinkreiser: So much `LOG_INFO`. i think mjolnir could use some log cleanup for sure. if this is really debug it would be best to `LOG_DEBUG` imho

kevinkreiser:
other than log comments ship

PR #243, opened 2015-05-27
set e so we dont ahve to or on exit. log when the lockfile is there. …

kevinkreiser:
…dont need to sed off the quotes jq has a switch for that. have shebang everywhere. clear all logs periodiacally

### DIFF
--- a/scripts/backup_tiles.sh
+++ b/scripts/backup_tiles.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-date_time=`date +%Y_%m_%d_%H%M%S`
-mv $1 $1.$date_time
-mkdir -p $1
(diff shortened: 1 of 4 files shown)


scripts/backup_tiles.sh
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
-
-date_time=`date +%Y_%m_%d_%H%M%S`
-mv $1 $1.$date_time
```
gknisely: where are we making backups now?

kevinkreiser: yep i just moved those two  lines into the main script. less scripts makes it easier to figure out what the heck is going on imho

gknisely: k

gknisely:
:ship:

PR #245, opened 2015-05-27
CHEFIFY ALL THE THINGS!!!!

kevinkreiser:
despite what it looks like, this is basically the same code just way less of it (therefore easier to debug)

update scripts were trapping the lock file which means that if one was interrupted it would never run again. 
removed seprate paths for with vs without updates and made it work with env var (since they were largely the same). 
added python script for moving tiles to s3 and telling service instances to get the updated tiles.
changed pbfgraphbuilder to log to stdout so it lands in the same file that we forward output to
clear all logs on a semi regular interval


kevinkreiser:
still need a script to pull down the tiles from s3, coming up next

gknisely:
:ship:

kevinkreiser:
i've moved all the scripts after testing they work into the templates portion of the chef repo (pr forthcoming). this allows us to template the pareamters to the scripts and in general its all stuff for chef anyway so it should belong there anyway. it also greatly simplifies how we use the things in chef land as well

PR #248, opened 2015-05-28
Graph Validator of Ultimate Speed mk. II

Kopkins:
Parallelized graphvalidator.cc to improve speed

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -13,6 +13,12 @@
 #include <vector>
 #include <map>
 #include <utility>
+#include <queue>
+#include <list>
+#include <thread>
+#include <future>
+#include <mutex>
+#include <numeric>
 
 #include <valhalla/midgard/logging.h>
 #include <valhalla/midgard/pointll.h>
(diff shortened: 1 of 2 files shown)


src/mjolnir/graphvalidator.cc
```diff
@@ -26,15 +32,59 @@ using namespace valhalla::midgard;
 using namespace valhalla::baldr;
 using namespace valhalla::mjolnir;
 
+class validator_stats {
```
kevinkreiser: should be in anonymous namespace or in its own file

src/mjolnir/graphvalidator.cc
```diff
@@ -99,75 +149,48 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
...
-    for (uint32_t tileid = 0; tileid < ntiles; tileid++) {
-      // Get the graph tile. Skip if no tile exists (common case)
-      const GraphTile tile(tile_hierarchy_, GraphId(tileid, level, 0));
-      if (tile.size() == 0) {
-        continue;
+    // Local Graphreader
+    GraphReader propreader(hierarchy_properties);
+
+    // Get some things we need throughout
+    lock.lock();
```
kevinkreiser: no lock needed here

src/mjolnir/graphvalidator.cc
```diff
@@ -99,75 +149,48 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
...
+    // Check for more tiles
+    while (true) {
+      lock.lock();
+      if (tilequeue.empty()) {
+        lock.unlock();
+        break;
       }
+      // Get the next tile Id
+      GraphId tile_id = tilequeue.front();
+      uint32_t id = tile_id.tileid();
```
kevinkreiser: is `id` an unused variable?

src/mjolnir/graphvalidator.cc
Kopkins: This should have been in the anonymous namespace. Should probably move it to it's own file if we use this class to gather the rest of the statistics Dave wants.

src/mjolnir/graphvalidator.cc
```diff
@@ -99,75 +149,48 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
...
+      tilequeue.pop();
+      lock.unlock();
 
-      // Iterate through the tile and validate signs
-      for (uint32_t i = 0; i < tile.header()->nodecount(); i++) {
-        const NodeInfo* nodeinfo = tile.node(i);
+      // Point tiles to the set we need for current level
+      int level = (int)tile_id.level();
```
kevinkreiser: no cstyle casts please. if you must cast use `static_cast`, id highly suggest switching this to be an auto though and making your levels vector be of uint32 or uint64 or whatever the type is

src/mjolnir/graphvalidator.cc
```diff
@@ -99,75 +149,48 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
...
-      // Get the graph tile. Skip if no tile exists (common case)
-      const GraphTile tile(tile_hierarchy_, GraphId(tileid, level, 0));
-      if (tile.size() == 0) {
-        continue;
+    // Local Graphreader
+    GraphReader propreader(hierarchy_properties);
+
+    // Get some things we need throughout
+    lock.lock();
+    auto tile_hierarchy = propreader.GetTileHierarchy();
```
kevinkreiser: `propreader` seems very misleading to me. elsewhere we name this `graph_reader` or `graphreader`

src/mjolnir/graphvalidator.cc
```diff
@@ -216,34 +249,113 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
         nodes.emplace_back(std::move(nodeinfo));
       }
 
-      float density = (roadlength * 0.0005f) / tiles.Area(tileid);
-      densities.push_back(density);
+      // Add density to stat class
+      float density = (roadlength * 0.0005f) / tiles->Area(id);
```
kevinkreiser: ah i found wher eyou used `id`, maybe could just call `tile_id.id()` instead of keeping another variable

src/mjolnir/graphvalidator.cc
```diff
@@ -216,34 +249,113 @@ void GraphValidator::Validate(const boost::property_tree::ptree& pt) {
...
+    std::random_shuffle(tempqueue.begin(), tempqueue.end());
+    std::queue<GraphId> tilequeue(tempqueue);
+    LOG_INFO("Done creating queue of tiles: count = " +
+             std::to_string(tilequeue.size()));
+
+    // An atomic object we can use to do the synchronization
+    std::mutex lock;
+
+    LOG_INFO("GraphValidator - Validating signs and connectivity");
+      // Spawn the threads
```
kevinkreiser: indentation is off

src/mjolnir/graphvalidator.cc
Kopkins: Was used one time. Changing to just have `tile_id.tileid()` in place of the variable

src/mjolnir/graphvalidator.cc
Kopkins: Can we use uint32 to index into a vector?

src/mjolnir/graphvalidator.cc
Kopkins: Thanks, when I merged the two loops in the original file there were two readers and I named one `propreader` since it only read the `hierarchy_properties`

kevinkreiser:
:racehorse:

PR #255, opened 2015-06-10
only ever open pbf files once

kevinkreiser:
this has the advantage that we hold onto the inodes of the files at the beginning of the program so that if an update process makes new files it wont affect us. this may be why running updates in the background while cutting tiles is not working. tests all pass, cut data from an extract had the exact same size after this change.


kevinkreiser:
sorry :cow2: :boy: have to see if this fixes asynchronous pbf update issue and it takes a while to test with the planet

kdiluca:
:+1:

PR #256, opened 2015-06-15
added wheelchair_boarding to stops table.  Also, added logic to suppo…

gknisely:
…rt wheelchair_accessible and bikes_allowed to trips and schedule.


dnesbitt61:
:shipit

PR #257, opened 2015-06-17
New statistic class in GraphValidator

Kopkins:
This class gathers statistics on the road lengths within tile and country boundaries. It collects the total road lengths for each road classification broken down by the tile ID and by the country ISO code. The statistics are logged to debug by default and the logging is part of the class. The class was also modified to handle the statistics of the tile levels that is output at the end of graph validation.

Some sample log output:
`2015/06/17 17:08:49.867722 [INFO] Country: US`
`2015/06/17 17:08:49.867740 [INFO]    Motorway: 3.00719e+07 Km`
`2015/06/17 17:08:49.867750 [INFO]    Primary: 2.75015e+07 Km`
`2015/06/17 17:08:49.867758 [INFO]    Residential: 1.85354e+08 Km`
`2015/06/17 17:08:49.867767 [INFO]    Secondary: 3.83092e+07 Km`
`2015/06/17 17:08:49.867774 [INFO]    ServiceOther: 2.53372e+07 Km`
`2015/06/17 17:08:49.867783 [INFO]    Tertiary: 3.95037e+07 Km`
`2015/06/17 17:08:49.867790 [INFO]    Trunk: 9.35517e+06 Km`
`2015/06/17 17:08:49.867799 [INFO]    Unclassified: 5.24041e+06 Km`

`2015/06/17 17:08:49.873237 [INFO] Tile: 749211`
`2015/06/17 17:08:49.873243 [INFO]    Motorway: 83180.5 Km`
`2015/06/17 17:08:49.873249 [INFO]    Primary: 42816.5 Km`
`2015/06/17 17:08:49.873256 [INFO]    Residential: 1.18913e+06 Km`
`2015/06/17 17:08:49.873263 [INFO]    Secondary: 124380 Km`
`2015/06/17 17:08:49.873269 [INFO]    ServiceOther: 126856 Km`
`2015/06/17 17:08:49.873276 [INFO]    Tertiary: 173731 Km`
`2015/06/17 17:08:49.873283 [INFO]    Trunk: 69589.8 Km`
`2015/06/17 17:08:49.873289 [INFO]    Unclassified: 22846.5 Km`

*The tile and country attributes are indented in the actual log*

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -19,7 +19,6 @@
 #include <future>
 #include <mutex>
 #include <numeric>
-#include <tuple>
 
 #include <valhalla/midgard/logging.h>
 #include <valhalla/midgard/pointll.h>
@@ -348,7 +347,7 @@ void validate(const boost::property_tree::ptree& hierarchy_properties,
       vStats.add_dup(dupcount, level);
     }
 
-    // Create return tuple and put into promise
+    // Fill promise with statistics
     result.set_value(vStats);
   }
 }
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -82,12 +82,12 @@ class validator_stats {
   std::map<std::string, std::map<RoadClass, float> > country_maps;
   std::set<uint32_t> tile_ids;
   std::set<std::string> iso_codes;
+
+public:
   const std::vector<RoadClass> rclasses = {RoadClass::kMotorway, RoadClass::kPrimary,
                                      RoadClass::kResidential, RoadClass::kSecondary,
                                      RoadClass::kServiceOther, RoadClass::kTertiary,
                                      RoadClass::kTrunk, RoadClass::kUnclassified};
-public:
-
 	validator_stats () : tile_maps(), country_maps(), iso_codes(), tile_ids() { }
 
 	void add_tile_road (const uint32_t& tile_id, const RoadClass& rclass, float length) {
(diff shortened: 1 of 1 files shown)


src/mjolnir/graphvalidator.cc
```diff
@@ -353,6 +417,35 @@ namespace mjolnir {
       LOG_INFO("Average density = " + std::to_string(average_density) +
                " max = " + std::to_string(max_density));
     }
+    // Whether or not to print ALL the stats
+    bool log_stats = false;
```
kevinkreiser: instead of setting the bool, you can just use `LOG_DEBUG` then the messages will only show on debug and lower builds

kevinkreiser:
is there any way we can combine `return_stats` and `validator_stats`?

src/mjolnir/graphvalidator.cc
```diff
@@ -353,6 +417,35 @@ namespace mjolnir {
...
+    if (log_stats) {
+      // Map to help print the roadway classifications
+      std::map<RoadClass, std::string> roadClassToString =
```
kevinkreiser: can we move this logging into the class itself? it seems the class will also eventually get a bunch of other functionality in terms of serializing itself, logging is one way, but outputing geojson is the next logical step. at that point we migth want to move the class out into its own seprate header and source file

kevinkreiser:
other than comments looks like a good first step. any idea if this impacts the run time (i would doubt it)? anyway :shipit: once  you've resolved the comments

dnesbitt61:
Agree about the logging - would be best if part of the class itself. A GeoJSON (or other file output) makes sense so we don't put all this stuff to one log file. We can discuss output fortats tomorrow.

PR #258, opened 2015-06-19
Nyc sf updates

gknisely:
We never used the no_thru_traffic flag for dest only.  Also, added shape key.


kevinkreiser:
whats shape_key do?

gknisely:
Shape_key is needed to query shape/shapes table.  :smile:

dnesbitt61:
:shipit:

PR #261, opened 2015-06-22
Update to statistics class

Kopkins:
Moved statistics class to its own file and added functionality to write the statistics to a DB file at /data/valhalla/statistics.sqlite.


src/mjolnir/graphvalidator.cc
```diff
@@ -1,17 +1,17 @@
...
-#include "valhalla/mjolnir/graphtilebuilder.h"
+#include "mjolnir/graphtilebuilder.h"
+#include "mjolnir/statistics.h"
 
 #include <valhalla/midgard/logging.h>
 
 #include <ostream>
-#include <set>
 #include <boost/format.hpp>
+#include <boost/filesystem/operations.hpp>
```
kevinkreiser: not needed i think

src/mjolnir/statistics.cc
```diff
@@ -0,0 +1,268 @@
...
+using namespace valhalla::mjolnir;
+
+namespace valhalla {
+namespace mjolnir {
+
+validator_stats::validator_stats ()
+  : tile_maps(), country_maps(), iso_codes(), tile_ids(), dupcounts(3), densities(3) { }
+
+void validator_stats::add_tile_road (const uint32_t& tile_id, const RoadClass& rclass, float length) {
+  this->tile_ids.insert(tile_id);
```
kevinkreiser: we haven't been using the `this` keyword at all in any of the other classes. i am personally not opposed to it (i secretely like it) but maybe for the sake of consistency we should remove it

src/mjolnir/graphvalidator.cc
Kopkins: was leftover from when the whole class was in that file.

dnesbitt61:
:shipit:

src/mjolnir/statistics.cc
```diff
@@ -0,0 +1,268 @@
...
+    LOG_ERROR("cannot open " + database);
+    sqlite3_close(db_handle);
+    db_handle = NULL;
+    return;
+  }
+
+  // Create table for tiles
+  sql = "CREATE TABLE tiledata (";
+  sql += "tileid INTEGER,";
+  sql += "motorway FLOAT,";
```
gknisely: Nothing major....but I usually follow the data types defined here:
https://www.sqlite.org/datatype3.html

Each column in an SQLite 3 database is assigned one of the following type affinities:

TEXT
NUMERIC
INTEGER
REAL
NONE

src/mjolnir/statistics.cc
```diff
@@ -0,0 +1,268 @@
...
+  ret = sqlite3_exec(db_handle, "BEGIN", NULL, NULL, &err_msg);
+  if (ret != SQLITE_OK) {
+    LOG_ERROR("Error: " + std::string(err_msg));
+    sqlite3_free(err_msg);
+    sqlite3_close(db_handle);
+  }
+  sql = "INSERT INTO tiledata (tileid, motorway, trunk, pmary, secondary, tertiary, unclassified, residential, serviceother) ";
+  sql += "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)";
+  ret = sqlite3_prepare_v2(db_handle, sql.c_str(), strlen (sql.c_str()), &stmt, NULL);
+  if (ret != SQLITE_OK) {
```
gknisely: Is ok to continue here or should we fail?

src/mjolnir/statistics.cc
```diff
@@ -0,0 +1,268 @@
...
+    LOG_ERROR("Error: " + std::string(err_msg));
+    sqlite3_free(err_msg);
+    sqlite3_close(db_handle);
+  }
+  sql = "INSERT INTO countrydata (isocode, motorway, trunk, pmary, secondary, tertiary, unclassified, residential, serviceother) ";
+  sql += "VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)";
+  ret = sqlite3_prepare_v2(db_handle, sql.c_str(), strlen (sql.c_str()), &stmt, NULL);
+  if (ret != SQLITE_OK) {
+    LOG_ERROR("SQL error: " + sql);
```
gknisely: Same...should we stop?

gknisely:
That is all I saw.  :+1:

src/mjolnir/statistics.cc
Kopkins: If we should return on error, we should add it to pbfadminbuilder too. Line 325

gknisely:
:shipit:

src/mjolnir/statistics.cc
dnesbitt61: I would never exit due to not being able to generate statistics. I can't imagine doing the whole planet successfully only to exit here. Log it is fine.

Kopkins: At the point where this code would execute the only remaining this is to write to the DB. Maybe we can review this tomorrow

PR #262, opened 2015-06-23
Ferry updates

gknisely:


src/mjolnir/pbfgraphparser.cc
```diff
@@ -503,9 +503,20 @@ struct graph_callback : public OSMPBF::Callback {
...
       w.set_speed(default_speed);
 
-// TODO  For now...drive on right.
+    //default to drive on right.
     w.set_drive_on_right(true);
 
+    // ferries / auto trains need to be set to highway cut off in config.
+    if (w.ferry() || w.rail()) {
+      RoadClass rc = w.road_class();
+      for (auto level : tile_hierarchy_.levels()) {
```
dgearhart: should level be a reference?

dnesbitt61: this value should be computed outside this (not every way callback)

dnesbitt61: though I guess if that is not feasible if you make this a reference it won't be too bad (since not tons of ferry ways)

gknisely:
@dnesbitt61 updated

dnesbitt61:
:shipit:  - Duane is right - should use a reference (auto &) but since this is now just an init step it is not crucial

gknisely:
@dgearhart @dnesbitt61 I missed @dgearhart  comment.  Updated.

dgearhart:
:+1:

PR #264, opened 2015-06-24
ReclassifyFerryConnections

dnesbitt61:
Reclassifiies edges so we get a highway path from roads to the ferry. Computes shortest path and considers driveability (to or from the ferry). Short ferries are skipped - they are likely to be river crossings where bridges are likely to be better alternatives.  Adds only 1-2 minutes to planet processing.

This fixes the Vancouver/Victoria ferry case and English Channel crossings (though not all have been tested).

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -6,6 +6,8 @@
 #include <utility>
 #include <thread>
 #include <set>
+#include <queue>
+#include <unordered_map>
 #include <boost/format.hpp>
 #include <boost/algorithm/string.hpp>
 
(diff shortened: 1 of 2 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -533,101 +542,338 @@ uint32_t GetBestNonFerryClass(const std::map<Edge, size_t>& edges) {
   return bestrc;
 }
 
+// NodeLabel - for simple shortest path
```
kevinkreiser: we should start moving massive bits of entact functionality out of graphbuilder and into their own files, just for sanities sake

dgearhart:
:shipit:

PR #265, opened 2015-06-24
Separate the reclassify link and ferry connection logic

dnesbitt61:
Moved into new files - this required moving the common node expansion code into its own file as well so it could be shared.


dgearhart:
:shipit:

kevinkreiser:
:racehorse:

PR #266, opened 2015-06-26
added transit shape.

gknisely:
May have to tweak after more testing.  But is a good start.

### DIFF
--- a/src/mjolnir/transitbuilder.cc
+++ b/src/mjolnir/transitbuilder.cc
@@ -576,14 +576,30 @@ std::vector<PointLL> GetShape(sqlite3* db_handle,
     auto start_idx = std::get<2>(start);
     auto end_idx = std::get<2>(end);
 
-    if (start_idx < end_idx) //forward direction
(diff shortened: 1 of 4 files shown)


src/mjolnir/transitbuilder.cc
```diff
@@ -537,18 +537,63 @@ Use GetTransitUse(const uint32_t rt) {
...
+      shape.push_back(stop_ll);
+      shape.push_back(endstop_ll);
+    }
+  } else {
+    shape.push_back(stop_ll);
+    shape.push_back(endstop_ll);
+  }
+
   return shape;
 }
```
dnesbitt61: Should the stop lls be part of the shape in the first 2 cases: going forward would be stop_ll, then the portion of the trip shape, then the endstop_ll? May want to try with/without this and see which looks better.

gknisely: Yes I did try this...I did not see a difference, but I will add back.  May need in the future for buses

dnesbitt61:
:shipit:

PR #272, opened 2015-06-30
Second update to statistic gathering

Kopkins:
The statistics are written to /data/valhalla/statistics.sqlite. The data is split between a few tables. There are total road lengths and road class lengths in the main tables for tiles and countries, located in tiledata and countrydata. Tile area is included as well.  

The other metrics now captured are:
* length of named roads
* length of one way roads
* length of roads with speed information
* number of roads/edges that are internal edges

This data is stored in the rclasstiledata and rclassctrydata tables. Each row in those tables has the mentioned data as well as the tile_id or the ISO code, along with the road classification, so for each tile or ISO there are 8 rows of data.

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -204,8 +204,8 @@ void validate(const boost::property_tree::ptree& hierarchy_properties,
             auto rclass = directededge.classification();
             tempLength /= (tileid == directededge.endnode().tileid()) ? 2 : 4;
             //Determine access for directed edge
-            auto fward = ((0x1 & directededge.forwardaccess()) == 0x1);
-            auto bward = ((0x1 & directededge.reverseaccess()) == 0x1);
+            auto fward = ((kAutoAccess & directededge.forwardaccess()) == kAutoAccess);
+            auto bward = ((kAutoAccess & directededge.reverseaccess()) == kAutoAccess);
             // Check if one way
             if ((!fward || !bward) && (fward || bward)) {
               vStats.add_tile_one_way(tileid, rclass, tempLength);
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -32,8 +32,6 @@ using namespace valhalla::mjolnir;
 
 namespace {
 
-
-
 // Get the GraphId of the opposing edge.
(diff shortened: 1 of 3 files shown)


src/mjolnir/graphvalidator.cc
```diff
@@ -201,11 +199,33 @@ void validate(const boost::property_tree::ptree& hierarchy_properties,
...
-          // Add road lengths to statistics class for current country and current tile
-          if (validLength) {
+          // Only consider edge if edge is good and it's not a link
+          if (validLength && !directededge.link()) {
             auto rclass = directededge.classification();
-            auto endnodeid = directededge.endnode().tileid();
-            tempLength /= (tileid == endnodeid) ? 2 : 4;
+            tempLength /= (tileid == directededge.endnode().tileid()) ? 2 : 4;
+            //Determine access for directed edge
+            auto fward = ((0x1 & directededge.forwardaccess()) == 0x1);
```
gknisely: should this be the following?
auto fward = (directededge.forwardaccess() & kAutoAccess) and
auto bward = (directededge.reverseaccess() & kAutoAccess);

Kopkins: It should definitely use kAutoAccess.

gknisely:
:ship after update for kAutoAccess

valhalla/mjolnir/statistics.h
```diff
@@ -25,11 +29,30 @@ namespace mjolnir {
...
+  // Length of named road
+  std::unordered_map<uint32_t, std::unordered_map<RoadClass, float, rclassHasher> > tile_named;
+  std::unordered_map<std::string, std::unordered_map<RoadClass, float, rclassHasher> > country_named;
+  std::unordered_set<uint32_t> tile_ids;
+  std::unordered_set<std::string> iso_codes;
+  std::unordered_map<uint32_t, float> tile_areas;
+  std::unordered_map<uint32_t, AABB2> tile_geometries;
   std::vector<std::vector<uint32_t> > dupcounts;
   std::vector<std::vector<float> > densities;
   std::map<RoadClass, std::string> roadClassToString =
```
kevinkreiser: these should be const static and not part of the class

kevinkreiser:
:rocket:

PR #277, opened 2015-07-08
Update to newer copy of ax_lua.m4.

bentley:
See https://www.gnu.org/software/autoconf-archive/ax_lua.html ;
from revision 5b82cc2556fdf7e32b99f99b64f25c6692ee1048 of
http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git

The version currently in mjölnir uses some non‐portable regular expressions that break on OpenBSD; this new version works.


kevinkreiser:
Thanks. Other repositories require it as well. Want to PR those too?

kevinkreiser:
merging will do the others

kevinkreiser:
turns out others that use lua probably dont need it! so yay for that

PR #278, opened 2015-07-08
Fix transit connection shape and length.

dnesbitt61:
<= is very important on the connection from the begin node, was skipping shape points.


gknisely:
:ship:

PR #285, opened 2015-07-29
fixed a bug where surface was never set when a user did not supply a surface tag.

gknisely:
before we were defaulting everything to kPavedSmooth if there was no surface tag.


dgearhart:
:shipit:

PR #286, opened 2015-07-30
set use = kMountainBike if is part of mountain bike network.

gknisely:


dgearhart:
:shipit:

src/mjolnir/directededgebuilder.cc
```diff
@@ -55,6 +54,10 @@ DirectedEdgeBuilder::DirectedEdgeBuilder(const OSMWay& way,
   else
     set_bikenetwork(way.bike_network());
 
+  if (attributes_.bikenetwork | kMcn)
+    set_use(Use::kMountainBike);
+  else set_use(use);
+
```
dnesbitt61: Set this in graphbuilder. Since use is passed into this method we shouldn't hide this behavior inside this method but make it expliclty in graphbuilder.

gknisely:
@dnesbitt61 good point.  updated.

dnesbitt61:
:shipit:

PR #287, opened 2015-07-30
Add Weighted Grade Factor to Edges

kevinkreiser:
~10-15% speed up with less shape access

the speed up is just for the initial tile writing phase. the rest of the process is unchanged in terms of performance (since i didnt touch it)

this is part 1 of a 2 part pr series to enable elevation querying at tile building time. the first step involves making sure we only ever look at shape when we absolutely need too. it turned out we were looking at the shape every time to get things like length (elevation and curvature to come) and the tiles that the shape cross (for connectivity map purposes). at any rate we now cache this information so that when we encounter the other directed_edge we look it up in a hash map.

in addition to all this i added the skadi dep to mjolnir so it can use it to get at elevation data

since it hasnt been merge we now have part 2. which actually adds the elevation querying. i'll do some testing to see how much this adds back to building initial tiles.

NOTE: this also contains a bunch of log clean up because i couldnt take it anymore


gknisely:
:mountain_bicyclist:

dnesbitt61:
:shipit

dnesbitt61:
:shipit:

PR #292, opened 2015-08-17
Fix shape issue

dnesbitt61:
Shape for shortcut edge was incorrect when the first edge had reverse direction shape. Improperly skipped the first shape point.


gknisely:
:ship:

PR #293, opened 2015-08-17
Name consistency - set to true if neither have any names.

dnesbitt61:


gknisely:
:ship:

src/mjolnir/graphenhancer.cc
```diff
@@ -905,6 +905,13 @@ bool ConsistentNames(const std::string& country_code,
...
   std::unique_ptr<StreetNames> street_names2 = StreetNamesFactory::Create(
       country_code, names2);
+
+  // Flag as consistent names when neither has names!
+  if (street_names1->empty() && street_names2->empty()) {
+    return true;
+  }
+
+  // Return true (consistent) if the common base names are not empty
   return (!(street_names1->FindCommonBaseNames(*street_names2)->empty()));
```
kevinkreiser: this shouldn't go into the FindCommonBaseNames function?

dgearhart: Find common base names is specific to named roads - i think the the unnamed case is separate

dgearhart:
:shipit:

src/mjolnir/graphenhancer.cc
kevinkreiser: fair enough sounds good to me

PR #295, opened 2015-08-21
Detection for edges that interfere with routing

Kopkins:
Writes out a JSON file that MapRoulette can use to /data/valhalla/tasks.json


src/mjolnir/statistics.cc
```diff
@@ -578,5 +578,64 @@ void validator_stats::build_db(const boost::property_tree::ptree& pt) {
...
+void validator_stats::RouletteData::AddTask (const PointLL& p, const uint64_t id, const std::vector<PointLL>& shape) {
+  auto result = way_IDs.insert(id);
+  if (result.second)
+    node_locs.insert({id, p});
+    way_shapes.insert({id, shape});
+}
+
+const std::unordered_map<uint64_t, PointLL> validator_stats::RouletteData::GetNodeLocs () const { return node_locs; }
+const std::unordered_map<uint64_t, std::vector<PointLL> > validator_stats::RouletteData::GetWayShapes () const { return way_shapes; }
+const std::set<uint64_t> validator_stats::RouletteData::GetWays () const { return way_IDs; }
```
kevinkreiser: this should return const std::set<uint64_t>& otherwise it is making a copy. also is there a reason that this needs to be sorted by id?

valhalla/mjolnir/statistics.h
```diff
@@ -141,6 +140,27 @@ class validator_stats {
...
+    std::unordered_map<uint64_t, PointLL> node_locs;
+    std::unordered_map<uint64_t, std::vector<PointLL> > way_shapes;
+    std::set<uint64_t> way_IDs;
+
+  public:
+    RouletteData ();
+
+    void AddTask (const PointLL& p, const uint64_t id, const std::vector<PointLL>& shape);
+
+    const std::unordered_map<uint64_t, PointLL> GetNodeLocs () const;
```
kevinkreiser: allow of these should return references

src/mjolnir/graphvalidator.cc
```diff
@@ -106,6 +108,144 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
...
+
+    if ((!(diredge->forwardaccess() & kAutoAccess) &&
+        (diredge->reverseaccess() & kAutoAccess)) ||
+        ((diredge->forwardaccess() & kAutoAccess) &&
+         (diredge->reverseaccess() & kAutoAccess)))
+      inbound++;
+  }
+
+  if ((outbound >= 2 && inbound == 0) || (inbound >= 2 && outbound == 0)) {
+    //std::cout <<  "Loop  " << tilebuilder.edgeinfo(directededge.edgeinfo_offset())->wayid() << std::endl;
```
gknisely: remove all the couts

src/mjolnir/statistics.cc
```diff
@@ -578,5 +578,64 @@ void validator_stats::build_db(const boost::property_tree::ptree& pt) {
...
+void validator_stats::RouletteData::Add (const RouletteData& rd) {
+  const auto new_ids = rd.GetWays();
+  const auto new_shapes = rd.GetWayShapes();
+  const auto new_nodes = rd.GetNodeLocs();
+  for (auto& id : new_ids) {
+    AddTask(new_nodes.at(id), id, new_shapes.at(id));
+  }
+}
+
+void validator_stats::RouletteData::GenerateTasks () {
```
kevinkreiser: we have a utility that lets you write json without all of this fun escaping and checking if you need a comma etc. you can find it in: https://github.com/valhalla/baldr/blob/master/valhalla/baldr/json.h

gknisely: My bad.  I should of pointed Kyle to that.  Sorry about that

PR #297, opened 2015-08-26
breakfix: shortcut edges being too distinct

kevinkreiser:
in my previous pr i ended up refactoring the shortcut building code to recompute the length of a given edge, this would lead to a more accurate (increasing in the number of contracted edges) estimation of a given edge's length but had the unfortunate side effect..

basically when it would compute the length of an edge the next time it would use the shape but in possibly a different direction. compounding trig functions and the use of only floating point precision means that if the value ended up being near x.5f the rounding could go one way or another depending which directed_edge of a given edge you were computing. anyway having the length be off by one meant that looking up the edge_info failed and so the two edges would not be considered directed_edges of the same edge.

what made this doubly tricky to find was that it required you to have admin info to expose it. must have had to do with the decision about whether or not two edges could contract based on matching admin info.

so i've reverted this to just add up the rounded lengths as before.

### DIFF
--- a/src/mjolnir/graphbuilder.cc
+++ b/src/mjolnir/graphbuilder.cc
@@ -359,9 +359,6 @@ void BuildTileSet(const std::string& ways_file, const std::string& way_nodes_fil
     std::map<GraphId, size_t>::const_iterator tile_end,
     std::promise<DataQuality>& result) {
 
-  std::string thread_id = static_cast<const std::ostringstream&>(std::ostringstream() << std::this_thread::get_id()).str();
-  LOG_INFO("Thread " + thread_id + " started");
-
   sequence<OSMWay> ways(ways_file, false);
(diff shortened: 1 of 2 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -359,9 +359,6 @@ void BuildTileSet(const std::string& ways_file, const std::string& way_nodes_fil
     std::map<GraphId, size_t>::const_iterator tile_end,
     std::promise<DataQuality>& result) {
 
-  std::string thread_id = static_cast<const std::ostringstream&>(std::ostringstream() << std::this_thread::get_id()).str();
```
kevinkreiser: the changes in this file are all just logging clean up i missed in the last round

gknisely:
:ship:

PR #299, opened 2015-09-01
fix grades

kevinkreiser:
yeah the grades that come out of skadi are now from -10 to 15 so we have to map that to a value of 0 to 15 which we can use as an index into a 16 value array of costing factors. too many last minute interface changes caused me to miss this.. apologies everyone


dnesbitt61:
:shipit

PR #300, opened 2015-09-03
loops be damned

kevinkreiser:
they are hard to work around so instead we just dont have them any more. having @dnesbitt61 test it against some known messed up cases


dnesbitt61:
Looks good in my testing. Didn't get exact number of the size increase in data but it was below 1% (probably well below)

kevinkreiser:
yeah seems less than a tenth of a percent at the most.

PR #301, opened 2015-09-08
consider full length and just end points of fairly short segments of road

kevinkreiser:
if an edge is less than 180m it will just use the end points to compute the weighted grade. this is important because the resampling doesnt sample the end point which for short edges means you can miss a significant portion of the edge.


dnesbitt61:
:shipit:

PR #304, opened 2015-09-21
kill loops forever. fixes #302

### LINKED ISSUE #302: no more loops! this time for real
so we made a fix not to allow loops that come from the ways, but it can turn out that you can get a loop that is at the graph edge level via a lolly pop way (a way that connects back to itself). 

the initial thing we did was detect when the first and last node of a way was the same. however we simply need a more generic rule. if any nodes of a way are the same then we have a loop and need to split it.

so the fix will be in the same place but be a little more costly (n^2 or nlogn) depending on implementation where n is the number of nodes on the way.

kevinkreiser:
we had originally fixed an easy case of loop where the begin and end of a way was the same node. now we split between any duplicate nodes along a given way (most commonly lolipop loops)

### DIFF
--- a/src/mjolnir/pbfgraphparser.cc
+++ b/src/mjolnir/pbfgraphparser.cc
@@ -203,6 +203,7 @@ struct graph_callback : public OSMPBF::Callback {
     last_way_ = osmid;
 
     // Add the refs to the reference list and mark the nodes that care about when processing nodes
+    loop_nodes_.clear();
     for (size_t i = 0; i < nodes.size(); ++i) {
       const auto& node = nodes[i];
       if(shape_.IsUsed(node)) {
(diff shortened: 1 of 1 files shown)


src/mjolnir/pbfgraphparser.cc
```diff
@@ -577,15 +583,9 @@ struct graph_callback : public OSMPBF::Callback {
...
-    // into two edges in the graph. If a problem is hard, avoid the problem!
-    if(nodes.front() == nodes.back()) {
-      intersection_.set(nodes[nodes.size()/2]);
-      //TODO: update osmdata_.*_count?
-      // Infer cul-de-sac if a road edge is a loop and is low classification.
-      if(w.use() == Use::kRoad && w.road_class() > RoadClass::kTertiary)
-        w.set_use(Use::kCuldesac);
-    }
+    // Infer cul-de-sac if a road edge is a loop and is low classification.
+    if(loop_nodes_.size() != nodes.size() && w.use() == Use::kRoad && w.road_class() > RoadClass::kTertiary)
```
gknisely: why excluding kResidential?

kevinkreiser: I'm just moving the cheese. ask @dnesbitt61 https://github.com/valhalla/mjolnir/commit/fb78cf394fc7cffa8ddd472981ce70b3734f0767#diff-482c3171d55aca66d8079ce81c75ac7eR879

dnesbitt61:
:shipit

PR #305, opened 2015-09-21
use lists to avoid reallocation

kevinkreiser:
there are places where we work with shape that we dont know its size apriori. there are also places were we add arbitrary bits of shape together. both of these things suck when using vectors because of reallocations. it turned out that none of this stuff required access by index either so no loss there.

this pr changes that stuff to use lists so that we can avoid reallocations. this made no difference in a small import like PA. we'll see if it makes a difference for larger imports.

the main reason i did this was because elevation sampling has no clue how many points it will end up generating so its hard to accurately preallocate space. this also benefits hierarchy building as it will `splice` arbitrary sections of road together. this operation is O(1) for lists and O(n) for vectors, plus vectors would be reallocating

ive also templated the methods to work with either lists or vectors depending on whats needed so as not to break stuff that does actually need `random_access_iterators`

### DIFF
--- a/src/mjolnir/edgeinfobuilder.cc
+++ b/src/mjolnir/edgeinfobuilder.cc
@@ -27,9 +27,13 @@ void EdgeInfoBuilder::AddNameOffset(const uint32_t offset) {
 }
 
 // Set the shape of the edge. Encode the vector of lat,lng to a string.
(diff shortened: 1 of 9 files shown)


gknisely:
:ship:

PR #307, opened 2015-09-29
Added verification so we do not exceed 15 names per edge

dgearhart:
@gknisely

### DIFF
--- a/src/mjolnir/edgeinfobuilder.cc
+++ b/src/mjolnir/edgeinfobuilder.cc
@@ -19,7 +19,7 @@ void EdgeInfoBuilder::set_wayid(const uint64_t wayid) {
 void EdgeInfoBuilder::set_text_name_offset_list(
     const std::vector<uint32_t>& text_name_offset_list) {
   if (text_name_offset_list.size() > kMaxNamesPerEdge) {
-    LOG_ERROR("Exceeding max names per edge: " +
+    LOG_WARN("Tried to exceed max names per edge: " +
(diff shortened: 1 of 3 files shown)


src/mjolnir/edgeinfobuilder.cc
```diff
@@ -18,12 +18,21 @@ void EdgeInfoBuilder::set_wayid(const uint64_t wayid) {
 // Set the indexes to names used by this edge.
 void EdgeInfoBuilder::set_text_name_offset_list(
     const std::vector<uint32_t>& text_name_offset_list) {
-  text_name_offset_list_ = text_name_offset_list;
+  if (text_name_offset_list.size() > kMaxNamesPerEdge) {
+    LOG_ERROR("Exceeding max names per edge: " +
```
kevinkreiser: shouldnt these be warns? i mean if we are going to continue the process? wouldnt error be something unrecoverable?

gknisely: Also, since you have a PR for mjolnir, can you change this one to warning as well?  When done let me know.  I am waiting for this PR and then I will update dev.
2015/09/29 14:23:38.693563 [ERROR] Exceeding max shortcut edges from a node: 8

dgearhart: i will change to LOG_WARN after current conversion is done

kevinkreiser:
ditto what greg said. :shipit:

PR #315, opened 2015-10-28
matches with the rest of the programs logging

kevinkreiser:
the other builders log more so in this manner. for the sake of symmetry and less log lines this seemed better


dnesbitt61:
:shipit:

PR #321, opened 2015-12-07
Transit and Tile Format Changes

kevinkreiser:
Transit updates to go along with new data formats. Removed "builders" and now rely on set methods in baldr classes.


dnesbitt61:
:shipit:

PR #325, opened 2015-12-09
get stop pairs reliably and log less 429s

kevinkreiser:
so now we get stop pairs by asking for all stop pairs of a particular stop in a particular tile. the question would be how many more queries would this take. lets do some math...

we have 26297 stop in the data which means probably that many queries for the new approach (although we do limit to a few thousand at a time, hopefully any given one will have less than 1k)

for the previous approach there were 2744357 not counting all the new ones recently added in NYC. generally we were asking for stop pairs in increments of 1k or 5k (its configurable). this would make it an order of magnitude less queries. so we need to see if we can put more than one origin for the above approach. both having the parameter twice and comma separated dont work so we should ask for this feature possibly.


gknisely:
:train:

kevinkreiser:
added a commit still ship?

dgearhart:
:+1:

PR #326, opened 2015-12-10
properly account for geodesics when fetching items into cartesian bounds

kevinkreiser:
basically we need to take the worst case on both the top and bottoms of our bounding boxes because if we dont we'll miss a sliver of stuff on one of the sides.


dnesbitt61:
:shipit:

PR #332, opened 2015-12-15
pick mid point between a loop

kevinkreiser:
not sure how we missed that regression.

note that for really complicated self intersections this can produce suboptimal results but we dont expect those very frequently.


dgearhart:
:shipit:

dnesbitt61:
:shipit

gknisely:
lol @dnesbitt61 

:shipit:

PR #334, opened 2016-01-04
Update GetOpposingEdge 

dnesbitt61:
to compare wayIDs and edgeinfo offset (or shape if endnodes are not in the same tile) to get correct opposing edge on local level. Not effective on arterial and highway hierarchies due to wayIds being set to 0.

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -40,6 +40,7 @@ bool ShapesMatch(const std::vector<PointLL>& shape1,
   }
 
   if (shape1.front() == shape2.front()) {
+    // Compare shape in forward direction
     auto iter1 = shape1.begin();
     auto iter2 = shape2.begin();
     for ( ; iter2 != shape2.end(); iter2++, iter1++) {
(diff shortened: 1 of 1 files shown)


src/mjolnir/graphvalidator.cc
```diff
@@ -33,45 +33,119 @@ using namespace valhalla::mjolnir;
...
+    auto iter1 = shape1.begin();
+    auto iter2 = shape2.begin();
+    for ( ; iter2 != shape2.end(); iter2++, iter1++) {
+      if (*iter1 != *iter2) {
+        return false;
+      }
+    }
+    return true;
+  } else if (shape1.front() == shape2.back()) {
+    int n = shape2.size() - 1;
```
kevinkreiser: you could use a reverse iterator `shape2.rbegin()` if you didnt want to have to use the extra multiplications going on in the `operator[]`

src/mjolnir/graphvalidator.cc
```diff
@@ -387,6 +464,11 @@ namespace mjolnir {
                " max = " + std::to_string(max_density));
     }
     stats.build_db(pt);
+
+    auto t1 = std::chrono::high_resolution_clock::now();
+    uint32_t msecs = std::chrono::duration_cast<std::chrono::seconds>(
+				  t1 - t0).count();
+    LOG_INFO("GraphValidator took " + std::to_string(msecs) + " sec");
```
kevinkreiser: the logs already have the time stamps in them, must we really add this? we should add it to all the major steps in mjolnir if so

kevinkreiser: looks like there are still tabs in that as well

src/mjolnir/graphvalidator.cc
```diff
@@ -33,45 +33,122 @@ using namespace valhalla::mjolnir;
...
+                  match = true;
+                }
+              }
+            }
+          }
+          if (match) {
+            // Check if multiple edges match - log any duplicates
+            if (opp_index != absurd_index) {
+              if (startnode.level() == 2) {
+                LOG_INFO("Potential duplicate: wayids " + std::to_string(wayid) +
```
kevinkreiser: Might this be better as a `WARN`?

kevinkreiser:
:rocket:

PR #339, opened 2016-01-08
Gk - restriction updates

gknisely:
GetAccessRestrictions function def changed...need to know what restrictions we are interested in.


dnesbitt61:
:shipit:

PR #344, opened 2016-01-12
more transit fixes

gknisely:
if sp.service_start_date() == sp.service_end_date() then clear dow mask. Fix for end_day. Was calculated wrong due to old start_date logic.


dgearhart:
:clock1:

dgearhart:
:clock2:

PR #345, opened 2016-01-12
text offset fix

gknisely:
headsign must be added below the continue otherwise text offset will be off when we reject a feed.


dgearhart:
:shipit:

PR #349, opened 2016-01-14
small changes to make thinner transit tiles. conf as make target

kevinkreiser:


src/mjolnir/transit_fetcher.cc
```diff
@@ -718,7 +713,8 @@ int main(int argc, char** argv) {
   curl_global_init(CURL_GLOBAL_DEFAULT);
 
   //go get information about what transit tiles we should be fetching
-  auto transit_tiles = which_tiles(pt);
+  //auto transit_tiles = which_tiles(pt);
```
gknisely: is this just for the one tile hack

gknisely:
:shipit:

PR #352, opened 2016-01-19
Transit.n tiles

gknisely:
transit fetcher and builder updates to save and read .pbf and pbf.n files.

Not sure why builder is such a :christmas_tree:   I just added a loop for the .n files.


src/mjolnir/transit_fetcher.cc
```diff
@@ -73,6 +73,7 @@ struct curler_t {
         //it should be 200 OK
         if(http_code == 200) {
           bool threw = false;
+          std::string x = result.str();
```
kevinkreiser: Delete this

PR #356, opened 2016-01-19
Fix density computations

dnesbitt61:
the bbox around the node position was using km rather than meters so nodes near the edge of a tile didn't properly find roads in nearby tiles. This led to radically different density for nodes very close to one another! Redoing this led to a need to rebalance the relative density computation. Also, added code to reduce speed for untagged motorway and trunk roads in urban areas. fixes issue #351


gknisely:
:+1:

PR #357, opened 2016-01-20
stop impact updates 

gknisely:
added IsCyclewayUturn so that they get assigned a stop impact of 7.  
Also, updated bestrc to Unclassified so that kUnclassified,  kResidential, and kServiceOther are grouped together for stop_impact logic.


gknisely:
@dnesbitt61 circle fixed and I updated this with the stop impact logic you request.  Please review.

dnesbitt61:
May want Uturn on cycleway logic to be either left or right after thinking more - lets discuss more. But you can :shipit: in the meantime

PR #358, opened 2016-01-20
Update stop impact from a turn channel

dnesbitt61:
(lower stop impact but do not make 0). Update tiles to be a const ref. Remove lock/unlock around getting the tile hierarchy (now a const ref). Move some variable declarations out of the lock/unlock code to be closer to where they are used.


gknisely:
:ship: after circle good

gknisely:
Circle good...but depends on midgard

PR #360, opened 2016-01-22
Hgv stats

gknisely:
Added truck stats.  hazmat and truck_route columns are lengths.   height, width, length, weight, and axle_load are counts.

I need to verify these counts and values are correct.  Doing that next.


dgearhart:
:shipit:

PR #370, opened 2016-01-27
log stats for links

gknisely:
Must log stats for links for truck routing.

Example image of all the truck data on *_links.

image


dgearhart:
:shipit:

PR #375, opened 2016-02-02
Separate stats into separate method. Add per tile density.

dnesbitt61:
Add a method to get the address  of a tile header that can be edited. This is for use in GraphValidator to set tile relative header values (e.g. density). In the builder constructor, read the existing header (if tile is non-empty) and make it changeable so we can add the tile relative values.


dnesbitt61:
This fixes issue #363

gknisely:
:bar_chart:

PR #377, opened 2016-02-03
route_stop_patterns

gknisely:
Added shape to transit.

note in transitbuilder - distances are saved to a vector.  A map stores the points and begin and end indexes and is looked up via a shape_id.

OLD:
image

NEW:
image

### DIFF
--- a/src/mjolnir/transit_fetcher.cc
+++ b/src/mjolnir/transit_fetcher.cc
@@ -420,6 +420,7 @@ bool get_stop_pairs(Transit& tile, unique_transit_t& uniques, const std::unorder
       }
     }
 
+    // shape data
     std::string shape_id = pair_pt.second.get<std::string>("route_stop_pattern_onestop_id", "null");
     if(shape_id != "null") {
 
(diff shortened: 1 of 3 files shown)


src/mjolnir/transit_fetcher.cc
```diff
@@ -506,6 +553,20 @@ void fetch_tiles(const ptree& pt, std::priority_queue<weighted_tile_t>& queue, u
       request = response.get_optional<std::string>("meta.next");
     } while(request && (request = *request + api_key));
 
+    std::unordered_map<std::string, size_t> shapes;
```
kevinkreiser: for the sake of symmetry itd be cool if you could put a comment above this block discribing what its fetching from the api

gknisely:
updated

src/mjolnir/transit_fetcher.cc
```diff
@@ -272,6 +273,29 @@ void get_routes(Transit& tile, std::unordered_map<std::string, uint64_t>& routes
...
+    auto* shape = tile.add_shapes();
+    auto shape_id = shape_pt.second.get<std::string>("onestop_id");
+
+    std::vector<PointLL> trip_shape;
+    PointLL  ll;
+    for(const auto& geom : shape_pt.second.get_child("geometry.coordinates")) {
+      auto lon = geom.second.front().second.get_value<float>();
+      auto lat = geom.second.back().second.get_value<float>();
+      ll.Set(lon,lat);
+      trip_shape.push_back(ll);
```
kevinkreiser: emplace_back (lon,lat) saves some copies

kevinkreiser:
Dank je wel

dnesbitt61:
:shipit:

PR #378, opened 2016-02-04
Add Edge Binning for Improved Search

kevinkreiser:
This adds edge binning to the validator step in `pbfgraphbuilder` (we should really change the name of that..) 

It makes use of a bunch of stuff in `baldr` and `midgard` and adds a test to confirm its working properly.

It turned out that one bug was just that not all in between tiles `tweeners_t` weren't being written in `graphbuilder.cc` (and rightly so) but for binning purposes we needed MOAR. So i've moved that to the validator.

~~Currently the test is passing and it cuts tiles fine without saying anything is wrong but routing is broken sometimes and works others. So I'm working on beefing up the test to catch the final bugs with the changed structure of the tiles.~~

The final bug that I found was that I was using a `GraphTile*` that didnt have the changes applied to it from the validator. This meant that the validators changes to nodes and edges werent getting stored, which makes me think that opposing edges were wrongly set and was borking the route computation. But thats all good now.

A run of North America shows an increase of data size in the sub 5% range.


src/mjolnir/dataquality.cc
```diff
@@ -28,11 +28,6 @@ void DataQuality::AddStatistics(const DataQuality& stats) {
   for (uint32_t i = 0; i < 128; i++) {
     node_counts[i] += stats.node_counts[i];
   }
-
```
kevinkreiser: all changes in here are removing blank tiles creation

src/mjolnir/graphbuilder.cc
```diff
@@ -333,52 +333,6 @@ uint32_t AddAccessRestrictions(const uint32_t edgeid, const uint64_t wayid,
   return modes;
 }
 
-// Walk the shape and look for any empty tiles that the shape intersects
```
kevinkreiser: all changes in here are removing blank tiles creation

src/mjolnir/graphtilebuilder.cc
```diff
@@ -45,7 +45,6 @@ GraphTileBuilder::GraphTileBuilder(const baldr::TileHierarchy& hierarchy,
     textlistbuilder_.emplace_back("");
     text_offset_map_.emplace("", 0);
     text_list_offset_ = 1;
-    write_bins_ = false;
```
kevinkreiser: instead of serializing and deserializing in the builder it seemed handy to implement the binning such that you can add more binned edges to a tile any time. this is done with a simple static function below

src/mjolnir/graphvalidator.cc
```diff
@@ -510,17 +600,24 @@ namespace mjolnir {
         // If tile exists add it to the queue
         GraphId tile_id(id, level, 0);
         if (GraphReader::DoesTileExist(hierarchy, tile_id)) {
-          tempqueue.push_back(tile_id);
+          tilequeue.emplace_back(std::move(tile_id));
         }
       }
     }
-    std::random_shuffle(tempqueue.begin(), tempqueue.end());
-    std::queue<GraphId> tilequeue(tempqueue);
```
dnesbitt61: still want to shuffle the queue to reduce chances of threads working on adjacent tiles.

dnesbitt61:
:shipit:

src/mjolnir/graphvalidator.cc
kevinkreiser: yep accidental removal when removing the copy, putting it back

PR #379, opened 2016-02-04
Fix bug in prior commit where header graphid is set to (0,0,0)

dnesbitt61:
when no current tile exists.


kevinkreiser:
could just forgo the else and always set it but either way works! :rocket:

Issue #310, opened 2015-10-23
Roundabouts are considerer as road


costales:
Hi!
In this roundabout, the Valhalla type is not a roundabout.

Screenshot:
![roundabout](https://cloud.githubusercontent.com/assets/5886786/10694197/25288bba-799d-11e5-82d6-d604765127c2.jpg)

URL to OSM:
https://www.openstreetmap.org/edit?editor=potlatch2#map=19/43.32618/-5.73105

Thanks a lot in advance!

dgearhart:
@costales - it appears the issue is because it is marked as a motorway_link and roundabout - looking to update logic to handle properly

dgearhart:
@costales - the roundabout fixes are in production

costales:
Hi! I think all works perfect now. If I see anything wrong I will tell you.
Thanks a lot for your support!!



-- 
Enviado con Dekko desde mi dispositivo Ubuntu

PR #382, opened 2016-02-05
Reduce stop impact if all links (ramps and turn channels).

dnesbitt61:
Fixes a NJ jug handle case (thor issue #310).


dnesbitt61:
Update to ReclassifyLinks to fix thor issue #310

dgearhart:
:shipit:

PR #390, opened 2016-02-16
temp shape fix?

gknisely:
hack to fix segfault in dev and prod.  Sent examples to transit team.  Only happens in CA


kevinkreiser:
:rocket:

PR #394, opened 2016-02-17
move lua from conf back into mjolnir, and make it so that its built i…

kevinkreiser:
…nto the library

eventually lua will bit the dust and we'll just use c++ to process tags. this is one step closer to that. now the lua program is an in memory string, which we load instead of a file somewhere on a disk. this removes a bunch of stuff from the config files as well.

the big change here is really that i combined edge and node luas into a single graph lua. it was really just code motion but still a bit nerve racking.

### DIFF
--- a/lua/graph.lua
+++ b/lua/graph.lua
@@ -1,0 +1,1269 @@
+--TODO: check if you can use lua type boolean instead of strings and pass that back to osm2pgsql
+--with the hopes that they will become strings once they get back to c++ and then just work in
+--postgres
+
+highway = {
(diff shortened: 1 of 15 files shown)


lua/graph.lua
```diff
@@ -0,0 +1,1269 @@
...
+["dismount"] = "true",
+["discouraged"] = "false",
+["forestry"] = "false",
+["destination"] = "true",
+["customers"] = "true",
+["official"] = "false",
+["public"] = "true",
+["restricted"] = "true",
+["allowed"] = "true",
+["emergency"] = "false"
```
kevinkreiser: this was broken in the original edges.lua

gknisely:
:ship: after NA passes

dnesbitt61:
:shipit:

PR #398, opened 2016-02-22
Do not update ramp seed if it is tagged

dnesbitt61:
Do not update ramp speed if it is tagged in OSM (fixes Leesburg Bypass routes that got off highway onto the ramp and then got back on).


gknisely:
:ship:

PR #399, opened 2016-02-23
Reduce speeds on ramps

dnesbitt61:
so they do not exceed main roadways. Add 1 to stop impact on merge from ramp to non-ramp. Skipping transit connect edges caused issues - still further investigation needed there.


src/mjolnir/graphenhancer.cc
```diff
@@ -932,6 +933,9 @@ uint32_t GetStopImpact(uint32_t from, uint32_t to,
   // (ramps and turn channels) are involved.
   if (allramps || edges[from].use() == Use::kTurnChannel) {
     stop_impact /= 2;
+  } else if (edges[from].use() == Use::kRamp && edges[to].use() != Use::kRamp) {
```
dgearhart: is this specific enough for merge? so we do not stop using turn lanes

dgearhart:
:shipit:

PR #401, opened 2016-02-24
dish of cereal

gknisely:
added support for cyclestreet and bicycle_road. 
Fixed ferry issue where access is turned off, but we still allowed routing.


dgearhart:
:minibus:

dnesbitt61:
:shipit:

Issue #331, opened 2015-12-11
Way doubles back on itself


dnesbitt61:
https://www.openstreetmap.org/way/11914359

Creates duplicate edges. Can we detect and identify for edits or automated fix?

PR #408, opened 2016-03-01
Add code to detect ways that loop back on themselves

dnesbitt61:
(node n == node n+2). Output wayid to file for later submission to a tool like MapRoulette to fix.


dgearhart:
:ship:

dnesbitt61:
closes #331

PR #412, opened 2016-03-02
official was not correct.

gknisely:


gknisely:
changed my mind on this one....turned on bus and foot but not motor_vehicle, bus, and truck.  they should probably be country level access tags.

PR #413, opened 2016-03-03
Ramp speed reduction

dnesbitt61:
Slight reduction of speed on untagged (no max_speed tag) ramps to make them slightly lower than roads of same category.  Fixes thor#327


dgearhart:
:shipit:

PR #415, opened 2016-03-08
Do not contract nodes where a gate or toll booth exist.

dnesbitt61:
Do not want shortcuts to go through toll booths and remove the ability to penalize tolls.
Fixes valhalla-docs#71


kevinkreiser:
:rocket:

PR #419, opened 2016-03-11
Transit connections - name consistency = false

dnesbitt61:
Make sure name consistency is false entering/exiting transit connections (they used to be unnamed) so that a maneuver penalty is triggered.


dgearhart:
:shipit:

PR #422, opened 2016-03-17
small shape update.

gknisely:
Transitland team clean up their shape but it mucked up the the upper_bound call.  

Fixes the following example:

screenshot from 2016-03-16 15 25 46


dgearhart:
:ship:

PR #423, opened 2016-03-22
Rework TransitDepartures

dnesbitt61:
Separate schedules from departures. Find unique schedule validity records and add to the tile. Replace transfers with schedules in tile builder. Use line Id as the  edge Id since now there can be more than 1 transit edge between 2 stops with the same route Id. Use headsign in addition to route Id and end stop when forming unique edges. This is so 2 different trips on the same route are separated when one goes to the end of the line and the other does not (for example).


gknisely:
:ship:

dnesbitt61:
Take out the change to use line Id for EdgeInfo lookup and also do not create separate directed edge using headsign. This caused some issues.

PR #424, opened 2016-03-22
move edge binning for testability

kevinkreiser:
`loki` has tests that create fake tiles and test correlating/search on them. what i want to do is move the code that creates this fake tile into `mjolnir` (there is an issue for that) but until we have the time to do that i just wanted expose the whole interface for making tiles so that `loki` and update its test tile so its tests pass with the new edge binning code.

so all this is really is moving a method from an anonymous namespace to a static class method


dnesbitt61:
:shipit:

PR #430, opened 2016-03-29
admin and name updates

gknisely:
Fix admin config issues.
Fix the broadway alt name issue.  Also, removed dups.

Old:
2: Turn left onto Broadway/St 1066.

New:
2: Turn left onto Broadway.

### DIFF
--- a/src/mjolnir/osmway.cc
+++ b/src/mjolnir/osmway.cc
@@ -560,6 +560,7 @@ bool OSMWay::link() const {
   return classification_.fields.link;
 }
 
+
 // Get the names for the edge info based on the road class.
 std::vector<std::string> OSMWay::GetNames(const std::string& ref,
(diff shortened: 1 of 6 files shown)


dgearhart:
:shipit:

src/mjolnir/osmway.cc
```diff
@@ -606,7 +607,7 @@ std::vector<std::string> OSMWay::GetNames(const std::string& ref,
   // Process name_en_
   // TODO: process country specific names
   if (name_en_index_ != 0)
-    names.emplace_back(ref_offset_map.name(name_en_index_));
+    names.emplace_back(name_offset_map.name(name_en_index_));
```
gknisely: @dgearhart this is the broadway fix

dgearhart:
would like to see some RAD results but if confident then :shipit:

PR #444, opened 2016-04-18
Pad EdgeInfo output (and SizeOf) to 8 byte boundaries

dnesbitt61:
so each way Id (uint64_t) starts on an 8 byte boundary.


kevinkreiser:
@dnesbitt61 whats the impact on tile size?

dgearhart:
:shipit:

dnesbitt61:
Looks like this adds a little over 2% to the size of the tiles.

kevinkreiser:
:rocket:

PR #445, opened 2016-04-20
Update Turn Channel Logic

dnesbitt61:


dgearhart:
Seem mostly good - I am still not sure what the deal is with the river exit
:shipit:

PR #449, opened 2016-05-12
Fetcher builder updates

gknisely:
Moved transit stops, schedules, shape, and transit_line logic to transit_fetcher.
Transit builder now just makes the transit connections from level 2 to 3 and vice versa.  

No diffs in NYC or SF routes.  Small diffs in Rome.  Actually, better routes.

### DIFF
--- a/src/mjolnir/transit_fetcher.cc
+++ b/src/mjolnir/transit_fetcher.cc
@@ -1281,10 +1281,6 @@ void AddToGraph(GraphTileBuilder& tilebuilder_transit,
     node.set_edge_index(tilebuilder_transit.directededges().size());
     node.set_timezone(stop.timezone());
 
-    bool admin_set = false;
-
-//TODO - admin lookup
-
  /** TODO - future when we get egress, station, platform hierarchy
     // Add any intra-station connections - these are always in the same tile?
     for (const auto& endstopid : stop_edges.second.intrastation) {
--- a/src/mjolnir/transit_fetcher.cc
+++ b/src/mjolnir/transit_fetcher.cc
@@ -23,14 +23,75 @@
 #include <valhalla/baldr/tilehierarchy.h>
 #include <valhalla/baldr/graphtile.h>
 #include <valhalla/baldr/datetime.h>
+#include <valhalla/baldr/graphreader.h>
 #include <valhalla/midgard/util.h>
 
+#include "mjolnir/graphtilebuilder.h"
 #include "proto/transit.pb.h"
 
 using namespace boost::property_tree;
 using namespace valhalla::midgard;
 using namespace valhalla::baldr;
 using namespace valhalla::mjolnir;
+
+// Shape
+struct Shape {
+  uint32_t begins;
+  uint32_t ends;
+  std::vector<PointLL> shape;
+};
+
+struct Departure {
+  GraphId  orig_pbf_graphid;   // GraphId in pbf tiles
+  GraphId  dest_pbf_graphid;   // GraphId in pbf tiles
+  uint32_t trip;
+  uint32_t route;
+  uint32_t blockid;
+  uint32_t shapeid;
+  uint32_t headsign_offset;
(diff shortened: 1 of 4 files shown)


src/mjolnir/transitbuilder.cc
```diff
@@ -15,7 +15,10 @@
 #include <spatialite.h>
 #include <fstream>
 #include <iostream>
+#define BOOST_NO_CXX11_SCOPED_ENUMS
```
kevinkreiser: this worries me. can you explain why we need it? we build valhalla with c++11 support but we get boost from debian packages which means its probably built without c++11 support (since its not defaulted on). mixing that kind of thing is asking for trouble.

src/mjolnir/transit_fetcher.cc
```diff
@@ -855,5 +1744,9 @@ int main(int argc, char** argv) {
   //spawn threads to connect dangling stop pairs to adjacent tiles' stops
   stitch(pt, all_tiles, dangling_tiles);
 
+  // update tile dir loc.  Don't want to overwrite the real transit tiles
+  if(argc > 4) { pt.get_child("mjolnir").erase("tile_dir"); pt.add("mjolnir.tile_dir", std::string(argv[4])); }
```
kevinkreiser: do you actually have to erase the value to put it? maybe that is the difference of `pt::add` vs `pt::put` just wondering, dont really know the answer

src/mjolnir/transitbuilder.cc
gknisely: was seeing the same error here...https://github.com/PurpleI2P/i2pd/issues/272

kevinkreiser:
how did you solve the issue of correlating to the route network using the osm way ids? you had said that since you only have the ids in the pbf tiles that you'd need access to the pbf again within pbfgraphbuilder to correlate them and even then you'd need ot have mapped them to the stops in the transit tiles. how did you end up making that work?

gknisely:
We grab the osm way ids from the stops in the transitbuilder.

kevinkreiser:
you're saying you keep the pbf around and indeed do access it in transitbuilder?

gknisely:
For now.  @dnesbitt61 wanted to start with that and then maybe change later if we come up with a solution.

kevinkreiser:
yeah i just wondered how it worked out. perfect stuff for the description of the pr :wink:

src/mjolnir/transit_fetcher.cc
gknisely: I just followed what u did for other args.

src/mjolnir/transitbuilder.cc
gknisely: without it I get this...

  CXXLD    transit_fetcher
../.libs/libvalhalla_mjolnir.so: undefined/.libs/libvalhalla_mjolnir.so reference: undefined to reference ` toboost: `:filesystemboost::::filesystemdetail::::copy_filedetail:(boost:copy_file::(boostfilesystem::::pathfilesystem: const:path&, const boost&, ::boost:filesystem::filesystem:path:: constpath &,const& boost, boost::filesystem::::copy_optionfilesystem:, :copy_optionboost:, :systemboost::::system::error_code*error_code*)')'

gknisely: make[1]: *** [pbfgraphbuilder] Error 1
./.libs/libvalhalla_mjolnir.so: undefined reference to `boost::filesystem::detail::copy_file(boost::filesystem::path const&, boost::filesystem::path const&, boost::filesystem::copy_option, boost::system::error_code*)'

kevinkreiser: yep i get that. the compiler doesnt like that style of enum, however if you include this file with one style and then include it elsewhere without that is bad news! so instead of hardcoding this define in the code we should pass it as a cflag to configure (like we do with the other boost preprocessor directives)

PR #450, opened 2016-05-13
Fix restrictions on shortcut edges

dnesbitt61:
Shortcut edges must set the restrictions field to the last directededge of the shortcut.


gknisely:
:ship:

PR #453, opened 2016-05-16
pass args to install

kevinkreiser:
this makes the configure optionally turn on coverage and there for make a debug build


gknisely:
:shipit:

PR #454, opened 2016-05-16
End tile fix

gknisely:
fixes the mass/boston core dump.

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -79,6 +79,10 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
 		uint64_t wayid, const GraphTile* tile, const GraphTile* end_tile,
 		uint32_t& dupcount, std::string& endnodeiso) {
 
+  if (!end_tile) {
+    LOG_ERROR("End tile invalid.")
+    return kMaxEdgesPerNode;
+  }
   // Get the tile at the end node and get the node info
   GraphId endnode = edge.endnode();
   const NodeInfo* nodeinfo = end_tile->node(endnode.id());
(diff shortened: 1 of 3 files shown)


src/mjolnir/transitbuilder.cc
```diff
@@ -379,18 +379,18 @@ void AddOSMConnection(const Transit_Stop& stop, const GraphTile* tile,
         if (directededge->use() == Use::kFerry || directededge->use() == Use::kRailFerry) {
           float start_distance = node->latlng().Distance(stop_ll);
 
-          const DirectedEdge* opp_de = tile->directededge(node->edge_index() + directededge->opp_index());
-          GraphId end_node = opp_de->endnode();
-          float end_distance = 0.0f;
-
           const GraphTile* endnode_tile = tile;
-          if ( directededge->endnode().Tile_Base() != end_node.Tile_Base()) {
+          if ( directededge->endnode().Tile_Base() != tile->id()) {
```
dnesbitt61: Tile_Base I think is tileid and level, so not sure about comparing to just id

gknisely: @dnesbitt61 Are you sure?  graphenhancer.cc line 405 and 1200 has this same logic.

dnesbitt61: On second look you are right. The tile->id() method returns the GraphId stored in the header, which should just include the tile Id and level.

gknisely: @dnesbitt61 ok.  Also added end_tile check in validator.

src/mjolnir/graphvalidator.cc
```diff
@@ -79,6 +79,10 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
 		uint64_t wayid, const GraphTile* tile, const GraphTile* end_tile,
 		uint32_t& dupcount, std::string& endnodeiso) {
 
+  if (!end_tile) {
```
dnesbitt61: Do you ever see this?

PR #455, opened 2016-05-17
High stop impact from turn channel onto ramp

dnesbitt61:


dgearhart:
When I tried to create north america data I got this:
terminate called after throwing an instance of 'std::runtime_error'
  what():  GraphTile NodeInfo index out of bounds: 762196,2,54181 nodecount= 9684
Aborted (core dumped)

dnesbitt61:
Updated to revert prior logic for stop impact. Also added changes to fix #457

dgearhart:
:shipit:

PR #461, opened 2016-05-24
New binary to build JSON used for data visualization

Kopkins:


valhalla/mjolnir/valhalla_build_statistics.h
```diff
@@ -0,0 +1,24 @@
+#include <string>
```
kevinkreiser: no need for this. if the functions are defined in the cc file and referenced no where else you may as well remove this file.

gknisely:
Please wait until production update is complete

PR #471, opened 2016-06-02
Updated stop impact logic for turn channel onto ramp (thanks @dnesbitt61)

dgearhart:
Updated the speed assignment for ramp to be a percentage of the original road class speed assignment

Closes valhalla/thor#338


gknisely:
:+1:

dnesbitt61:
:shipit:

PR #472, opened 2016-06-02
actually build all of the levels in geojson too

kevinkreiser:
a slight oversight lead to only outputting ppm for each level of the graph. now we also do geojson too instead of just level `2` 4 times


dgearhart:
:shipit:

dnesbitt61:
:shipit:

PR #474, opened 2016-06-03
toss proposed construction

gknisely:
If a user entered access tags, we would keep construction or proposed ways.

Updated results:

image


dnesbitt61:
:shipit:

PR #475, opened 2016-06-06
dont mix sizes to allow building on 32bit machines

kevinkreiser:
32bit machines fail builds when you assume that size_t is 64bits wide. another reason to not specify the number of bits you need unless you must to ensure wire compatibility, reduce memory, etc


gknisely:
:ship:

PR #478, opened 2016-06-07
DB update and json builder changes

Kopkins:
Changed db so that the order of road types is more logical

Changed a field in the json output to work better with other data in the valhalla-charts


dnesbitt61:
Will defer to Greg and Kevin - seems fine to me

gknisely:
:ship:

PR #480, opened 2016-06-08
Fix bugs in ReclassifyLinks

dnesbitt61:
Also add turn channel logic into ReclassifyLinks.


dgearhart:
it is a good improvement - thanks :shipit:

PR #483, opened 2016-06-08
Admin access

### LINKED ISSUE #452: country specific access logic

### LINKED ISSUE #479: Utrecht is too large for launchpad repositories
Currently we cant submit our full code to PPA and run tests there because the utrecht pbf is 30MB. We need to figure out the parts that we are testing there and extract them as separate smaller files. @gknisely i dont suppose you know the more exact spots you were testing? or would we have to look at each id and where it is to do the extracts?

gknisely:
Logic to support country specific access.  Based on country tables found here:  http://wiki.openstreetmap.org/wiki/OSM_tags_for_routing/Access-Restrictions

Basically, we now save if a user entered tags for auto, truck, bus, foot, and bike in access.bin.  (Note: If no tags were entered by the user, nothing is wrote out to access.bin.  Next, during the graphenhancer phase we determine if there are country overrides and update the forward and reverse access based on the country record stored in the admin db.  We do not update if a user entered a tag.  In other words, user inputs win over country specific access. 

The only updates that are currently stored in the db are trunk, trunk_link, track, footway, pedestrian, bridleway, cycleway, and path.  The reason for this is the fact that all other highway tags and uses do not differ from the defaults.

Note:  added tests and found and fixed other bugs too.  

fixes #479 
fixes #452 
fixes valhalla/thor#277

before:

image
     
after:

image


gknisely:
Depends on baldr updates

src/mjolnir/osmaccess.cc
```diff
@@ -0,0 +1,81 @@
+#include "mjolnir/osmaccess.h"
```
kevinkreiser: if there is one `OSMAccess` object per way it would be way cheaper to just add this access information to the `OSMway` class and store it in that sequence. if this is somehow a sparse set then separating is better.

valhalla/mjolnir/graphbuilder.h
```diff
@@ -25,8 +25,8 @@ class GraphBuilder {
    * and configs
    * @param  config         properties file
    * @param  osmdata        OSM data used to build the graph.
-   * @param  ways_file      where to store the ways so they arent in memory
-   * @param  way_nodes_file where to store the nodes so they arent in memory
+   * @param  ways_file      where to store the ways so they are not in memory
```
kevinkreiser: wow now who's the linguist pedant!?

src/mjolnir/osmaccess.cc
gknisely: There is not.  If no tags are entered by the user then I don't save.  Plus osm access is smaller

valhalla/mjolnir/graphbuilder.h
gknisely: you hate single quotes....I saw it a few times. :-)

kevinkreiser: :smile:

dgearhart:
:shipit: when tests pass on this :christmas_tree:

dnesbitt61:
:shipit:

test/graphparser.cc
```diff
@@ -45,9 +46,11 @@ void BollardsGatesAndAccess(const std::string& config_file) {
   boost::property_tree::ptree conf;
   boost::property_tree::json_parser::read_json(config_file, conf);
 
-  std::string ways_file = "test_ways.bin";
-  std::string way_nodes_file = "test_way_nodes.bin";
-  auto osmdata = PBFGraphParser::Parse(conf.get_child("mjolnir"), {"test/data/liechtenstein-latest.osm.pbf"}, ways_file, way_nodes_file);
+  std::string ways_file = "test_ways_bollards.bin";
```
kevinkreiser: every single test in this program doesnt have to have different names for the bin files. they dont run in parallel. only between utrecht.cc graphparser.cc and countryaccess.cc do they have to be different. inside of each test program they can use the same names.

gknisely: ok thanks

PR #484, opened 2016-06-14
Admin access

gknisely:
We now only process country access logic if the iso country codes match for the begin and end node.  A lot of moving stuff around.

### DIFF
--- a/src/mjolnir/countryaccess.cc
+++ b/src/mjolnir/countryaccess.cc
@@ -50,6 +50,7 @@ uint32_t GetAccess(const uint32_t current_access, const uint32_t country_access,
   return new_access;
 }
 
+// Set the country access for a DE.
 void CountryAccess(DirectedEdge& directededge, const std::vector<int> country_access,
                    const OSMAccess user_access) {
 
--- a/src/mjolnir/countryaccess.cc
+++ b/src/mjolnir/countryaccess.cc
@@ -51,8 +51,8 @@ uint32_t GetAccess(const uint32_t current_access, const uint32_t country_access,
 }
 
 // Set the country access for a DE.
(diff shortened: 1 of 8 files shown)


src/mjolnir/graphbuilder.cc
```diff
@@ -340,13 +341,76 @@ void BuildTileSet(const std::string& ways_file, const std::string& way_nodes_fil
...
     const uint32_t tile_creation_date,
+    const boost::property_tree::ptree& pt,
     std::promise<DataQuality>& result) {
 
   sequence<OSMWay> ways(ways_file, false);
   sequence<OSMWayNode> way_nodes(way_nodes_file, false);
   sequence<Edge> edges(edges_file, false);
   sequence<Node> nodes(nodes_file, false);
 
+  auto database = pt.get_optional<std::string>("admin");
```
kevinkreiser: would this chunk of stuff be a good candidate to move to `access.cc` as well?

src/mjolnir/countryaccess.cc
```diff
@@ -50,6 +50,7 @@ uint32_t GetAccess(const uint32_t current_access, const uint32_t country_access,
   return new_access;
 }
 
+// Set the country access for a DE.
 void CountryAccess(DirectedEdge& directededge, const std::vector<int> country_access,
```
kevinkreiser: it would be could if we'd name this with the word `set` otherwise its very confusing to read in the code

gknisely: sure np

src/mjolnir/graphbuilder.cc
gknisely: done

kevinkreiser:
looks good to me :rocket:

PR #485, opened 2016-06-16
The dead has been raised

Kopkins:
Pulled out code from my ancient pr #295 

This takes care of most of the changes from that pr


kevinkreiser:
i would love if we could take all of this stuff out of validator and put it into its own app. `valhalla_build_statistics` or something. that would separate the concerns of the two and allow it to be optional all in on shot. anyway that would be a job for another day if we would want to do it

Kopkins:
I agree. I really don't like jamming so much stuff into validator

PR #486, opened 2016-06-20
Remove reduction of stopimpact for cycleway --> cycleway transition

dnesbitt61:
This reduced the cost to cross higher class roads, which usually does require
stopping (with time penalty to slow down, stop, then accelerate again).


gknisely:
:ship:

PR #489, opened 2016-06-23
Pulled out database creation into its own binary

Kopkins:
To better organize and simplify things before starting the signage statistics, I pulled out the database creation into its own binary. That is now done in `valhalla_build_statistics` and there is no longer any statistics gathering in `valhalla_build_tiles`.

Additionally I moved the json generation for `Valhalla-Charts` into `statistics_json.cc` and it is not currently in the makefile. I'm trying to decide whether or not it should be in mjolnir since it is an independent tool used for things outside this repository.

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -256,7 +256,7 @@ void validate(const boost::property_tree::ptree& pt,
     GraphReader graph_reader(pt.get_child("mjolnir"));
     // Get some things we need throughout
     const auto& hierarchy = graph_reader.GetTileHierarchy();
-    auto numLevels = heirarchy.levels().size();
+    auto numLevels = hierarchy.levels().size();
     // vector to hold densities for each level
     std::vector<std::vector<float>> densities(numLevels);
     // Array to hold duplicates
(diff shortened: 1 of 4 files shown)


dnesbitt61:
:shipit:

src/mjolnir/graphvalidator.cc
```diff
@@ -794,14 +545,19 @@ namespace mjolnir {
...
+    std::vector<uint32_t> duplicates(3, 0);
+    std::vector<std::vector<float>> densities(3);
     tweeners_t tweeners;
     for (auto& result : results) {
-      auto pair = result.get_future().get();
-      //keep track of stats
-      stats.add(pair.first);
+      auto data = result.get_future().get();
+      // Total up duplicates for each level
+      for (uint8_t i = 0; i < 3; ++i) {
```
kevinkreiser: this shouldnt be hardcoded

src/mjolnir/graphvalidator.cc
```diff
@@ -794,14 +545,19 @@ namespace mjolnir {
     for (auto& thread : threads)
       thread->join();
     // Get the promise from the future
-    validator_stats stats;
+    std::vector<uint32_t> duplicates(3, 0);
```
kevinkreiser: similar here, get the number of levels from `tile_hierarchy`

PR #491, opened 2016-06-23
Simple program to create a real time speed file 

dnesbitt61:
using a csv input of way Id vs. speed. Used in a simple proof of concept to validate that tiled speed data can be used to influence routes and estimated times.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -131,7 +131,7 @@ bin_PROGRAMS = \
 	valhalla_query_transit \
 	valhalla_ways_to_edges \
 	valhalla_build_speeds \
-        valhalla_build_statistics
+	valhalla_build_statistics
 
 valhalla_benchmark_admins_SOURCES = src/mjolnir/valhalla_benchmark_admins.cc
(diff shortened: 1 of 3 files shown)


Makefile.am
```diff
@@ -130,7 +130,8 @@ bin_PROGRAMS = \
 	valhalla_build_transit \
 	valhalla_query_transit \
 	valhalla_ways_to_edges \
-  valhalla_build_statistics
+	valhalla_build_speeds \
+        valhalla_build_statistics
```
kevinkreiser: this should be tabs not spaces

Makefile.am
```diff
@@ -130,7 +130,8 @@ bin_PROGRAMS = \
 	valhalla_build_transit \
 	valhalla_query_transit \
 	valhalla_ways_to_edges \
-  valhalla_build_statistics
+	valhalla_build_speeds \
```
kevinkreiser: i like the name i think simple is the best approach in naming actual programs that will be run

kevinkreiser:
for larger inputs we'll not be able to use an in memory mapping but for now it looks ok for me

PR #494, opened 2016-07-07
Fix to not generate shortcuts that might drop certain maneuvers.

dnesbitt61:
Do not create shortcut at a node where the continuation along the same named road might be at a 'T' intersection or a turn in presence of other driveable roads. Creating a shortcut across these nodes will remove maneuvers that should be called out in the guidance/narrative.

Also, add an application to create real-time speed tiles from way Id and speed CSV files. This was used in a traffic proof of concept.


dgearhart:
I will run conversion tonight so I can run some tests tomorrow

dgearhart:
ok for now - if we see issues in the future we might have to improve the intersecting forward edge logic
:shipit:

Issue #492, opened 2016-06-24
Add support for 'no_turn' restriction


noblige:
Although not explicitly listed, I understand that `no_turn` is a valid restriction in relation (without specifying if it is left/right).

http://wiki.openstreetmap.org/wiki/Relation:restriction

noblige:
@dnesbitt61 Only change requested is to support `no_turn` in parser, there's no difference in relation. It's still same semantics - from/via/to, but instead of explicitly specifying `no_left_turn` or `no_right_turn` it's just `no_turn`.
It does not seems like it makes any difference - https://github.com/valhalla/mjolnir/blob/master/src/mjolnir/graphbuilder.cc#L277

PR #495, opened 2016-07-08
Restriction updates

gknisely:
#492 fixes 492


gknisely:
depends on baldr updates

dnesbitt61:
:shipit:

PR #497, opened 2016-07-11
Exit/Fork Statistics + MapRoulette Code Cleanup

Kopkins:
There are 4 new tables in statistics.sqlite:
* `tile_exitinfo`
* `tile_forkinfo`
* `ctry_exitinfo`
* `ctry_forkinfo`

These tables contain the id of the country or tile and the percent of forks or exits that have signage.

Additionally, I cleaned up the code for the Map Roulette task generation so that it now uses `baldr::json` and gets the file path from `conf/valhalla.json`.


kevinkreiser:
:+1:

PR #504, opened 2016-07-25
Do not check if admin string is empty in CanContract

dnesbitt61:
That prevents all shortcuts in cases where no admin DB is used.


dgearhart:
:ship:

PR #506, opened 2016-08-09
Initial Implementation of some Map Roulette Administration/Automation tools

Kopkins:
Create a local database by running `python3 py/challenge_admin/models.py`

Then you can interact with challenges by using `py/challenge_admin/admin_tool.py` 

This requires that you update `py/challenge_admin/config.py` with the proper api-key from the server you would like to access, as well as updating the server URL if not running a local instance of Map Roulette.


py/challenge_admin/admin_tool.py
```diff
@@ -0,0 +1,127 @@
...
+                            [{
+                                'type': 'Feature',
+                                'geometry': feature['geometry'],
+                                'properties': {}
+                            }]
+                        }
+                    }
+
+            # change some values depending on what type of task we built
+            if feature['properties']['type'] == 'Loop':
```
kevinkreiser: id like to keep all the generation stuff in c++ if we can. to that end could we use the `feature->properties->type` as a key into a list of instructions stored on the `featurecollection`? so in other words:

```json
{
  "type":"FeatureCollection",
  "features":[
    { ...,
      "properties":{"type":"loop"},
      ...
    }
  ],
  "properties": {
    "instructions": {
      "loop": "fix the loops punk",
      "dead_end": "fix the dead ends would you please"
    }
  }
}
```

py/challenge_admin/config.py
```diff
@@ -0,0 +1,8 @@
+
+database_file = 'sqlite:///maproulette.sqlite'
```
kevinkreiser: can we add the config to valhalla.json

src/mjolnir/statistics.cc
```diff
@@ -354,6 +354,16 @@ void statistics::RouletteData::GenerateTasks (const boost::property_tree::ptree&
   }
   // Add all unroutable nodes to the json array
   for (auto it = unroutable_nodes.cbegin(); it != unroutable_nodes.cend(); it++) {
+    uint64_t hash;
+    float* lat = new float;
+    float* lng = new float;
+    *lat = it->lat();
+    *lng = it->lng();
+    if (*lat > *lng) {
+      hash = *(reinterpret_cast<uint64_t*>(lat))<<32 | *(reinterpret_cast<uint64_t*>(lng));
```
kevinkreiser: https://github.com/valhalla/midgard/blob/master/valhalla/midgard/pointll.h#L154 should be useful here.

should be able to just:

```c++
auto hash = std::hash()(some_point);
```

although if you are doing it a lot you might want to keep the hash around instead of instantiating a new one every time

py/challenge_admin/config.py
Kopkins: @kevinkreiser do you want the geojson file to be specified in the config or as a parameter?

PR #507, opened 2016-08-10
Set flag for dead end directed edge

dnesbitt61:
This is based on the node intersection typebeing kDeadEnd. May need to revisit the setting of the node to dead end to consider if only 1 driveable edge exists.


src/mjolnir/graphvalidator.cc
```diff
@@ -92,6 +93,9 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
   // Set the end node iso.  Used for country crossings.
   endnodeiso = end_tile->admin(nodeinfo->admin_index())->country_iso();
 
+  // Set the deadend flag (TODO - do we need to worry about driveability)
+  deadend = nodeinfo->intersection() == IntersectionType::kDeadEnd;
```
dgearhart: i like parentheses around the equal check - just my opinion

dgearhart:
:shipit:

PR #509, opened 2016-08-19
Added GetOffsetForHeading method that will adjust heading offset based on road class and 'use'

dgearhart:


dnesbitt61:
:shipit:

PR #517, opened 2016-09-02
json only allows 64bit wide ints this fixes 32bit builds

kevinkreiser:
32bit builds were broken because of this


dgearhart:
:ship:

PR #531, opened 2016-09-22
access updates for transit. 

gknisely:
Now using the access from the transit tiles.  We set access at the node level via valhalla_build_transit and propagate it to the transit connections during the transitbuilder phase.  Finally, the access for transit stops is reset to (kPedestrianAccess | kWheelchairAccess | kBicycleAccess)


dnesbitt61:
:shipit:

PR #541, opened 2016-10-05
more sidewalk logic

gknisely:
do not set use to sidewalk for sidewalk kv.
added support for sidewalk use.


dnesbitt61:
:shipit:

PR #542, opened 2016-10-06
Separate shortcut building from hierarchy builder

dnesbitt61:


kevinkreiser:
lots of change with no description!? my brain is mush.. is there anything important to know in here other than shortcuts got moved out of hierarchy building and we're building them on the fly without precomputing which nodes are 'contractable'? looks to me like not but just wanted to ask.

:rocket:

src/mjolnir/hierarchybuilder.cc
```diff
@@ -584,8 +70,9 @@ std::pair<uint32_t, uint32_t> FormTilesInNewLevel(
     }
 
     // Check if we need to clear the tile cache
-    if(info.graphreader_.OverCommitted())
+    if (info.graphreader_.OverCommitted()) {
```
gknisely: if (meh) {
sigh {} for one line
}

gknisely:
:shipit:

dnesbitt61:
Separating the logic for building shortcuts simplifies the code within HierarchyBuilder. Also this now builds shortcuts on the fly, rather than gathering a list of nodes that can be contracted. This is still pretty fast, and simplifies the shortcut building slightly.

PR #543, opened 2016-10-07
make the edge bin tests less brittle

kevinkreiser:
@gknisely was working on restriction stuff and was noticing that this test was failing due to tile changes. this is an attempt to make this just as rigorous but far less brittle. also adds instructions for what to do when the tile format changes.


gknisely:
:+1:

PR #544, opened 2016-10-12
bin the other levels now that they arent all on the local

kevinkreiser:
this was a simple addition. all i had to do was remove the bit about skipping the other levels. after that i had to change the bit about checking tile ids based on edge id back to ids based on shape points. from there there was only one more change which was to make all the non local tiles bins look like `tweeners` so that they'd get added to the local levels at a later point in the validator.


src/mjolnir/graphtilebuilder.cc
```diff
@@ -668,9 +668,9 @@ void GraphTileBuilder::AddTileCreationDate(const uint32_t tile_creation_date) {
 using tweeners_t = std::unordered_map<GraphId, std::array<std::vector<GraphId>, kBinCount> >;
```
kevinkreiser: this file here has the only real changes to make this happen. changes in other files were simply cosmetic

src/mjolnir/graphvalidator.cc
```diff
@@ -471,20 +469,16 @@ void validate(const boost::property_tree::ptree& pt,
       ++start;
       lock.unlock();
 
-      //ignore transit tiles.
-      if (tile_bin.first.level() <= hierarchy.levels().rbegin()->second.level) {
```
kevinkreiser: this `if` isnt needed because we'll never have bins for any level but the local one

dgearhart:
:ship:

dnesbitt61:
:shipit:

PR #551, opened 2016-10-24
fixed transit crash

gknisely:
tile does not dangle if there are no stop pairs.  Added new vehicle types bus_service shuttle_bus.


dgearhart:
:ship:

PR #552, opened 2016-10-26
make tiles more robust to future needs without breakeage

kevinkreiser:
this is a corollary of: https://github.com/valhalla/baldr/pull/309


dnesbitt61:
:shipit:

PR #556, opened 2016-11-03
add try catch around TransitDeparture 

gknisely:
put a try catch around TransitDeparture constructor since we are doing a lot of checks now.


kevinkreiser:
:rocket:

PR #558, opened 2016-11-03
Make sure transfercount is set to 0 in the headerbuilder. 

dnesbitt61:
Add a placeholder for when we add transit transfer records.
Fixes countryaccess test.


gknisely:
:zero:

PR #560, opened 2016-11-04
Identify unused text strings when serializing a tile for updates. 

dnesbitt61:
Make sure they get properly added to the textlistbuilder (since we are not updating offsets of objects that may reference strings past that one in the list).


gknisely:
:ship:

PR #563, opened 2016-11-11
parse and store the largest changeset id you find

kevinkreiser:
this hangs the changeset id off of the `OSMdata` object. ~~ill make another commit to actually set it in the tile header and wont merge until thats done.~~

this worked perfectly for geofabrik extracts. i need to test with planet mirror and .osm converted to .pbf via osmconvert, i suspect they will be fine as well, we'll see


kevinkreiser:
:cow: :boy:

PR #564, opened 2016-11-11
Complex res updates

gknisely:
Added logic to store start restrictions on the "to" tile.  Also, we store end restrictions on the "from" tile.  The tests found this issue when the from or to is on different levels.  We got lucky before where the from and to were in the same tile.  All cr tests pass.


gknisely:
approved by @dnesbitt61

PR #565, opened 2016-11-18
properly handle maxspeed:forward/backward for one way streets

noblige:
old code was conflicting with logic in https://github.com/valhalla/mjolnir/blob/master/src/mjolnir/pbfgraphparser.cc#L713 and https://github.com/valhalla/mjolnir/blob/master/src/mjolnir/pbfgraphparser.cc#L708 resulting in speed set to `0` for one-way streets that had only either `maxspeed:forward` or `maxspeed:backward` tag set


noblige:
cc @kevinkreiser @dnesbitt61

PR #566, opened 2016-11-22
bins were missing some edges!

kevinkreiser:
fixes valhalla/loki#163 and fixes valhalla/meili#58

essentially there were two problems. number one was that the condition for binning a tweener shouldnt have been based off of the intersected tweener tile but rather the start and end tile. the other problem occured as a result of binning other levels of tiles into the level 2 tiles. we were using edgeinfo offset as a determination if we had binned an edge or not skipping edgeinfo offsets that we've already seen in a given tile. level 0 tiles though can contain two directed edges that start and end in two different level 2 tiles. so the edgeinfo offset is not unique enough to say that an edge is entirely contained within a level 2 tile. instead you must also check that the start and end level 2 tiles are the same and that you've seen that edgeinfo offset before.

whew! :smile: 

@dnesbitt61 @noblige @yzadik @smwork


src/mjolnir/graphtilebuilder.cc
```diff
@@ -721,33 +721,33 @@ std::array<std::vector<GraphId>, kBinCount> GraphTileBuilder::BinEdges(const Til
...
+    auto intersection = tiles.Intersect(shape);
     GraphId edge_id(tile->header()->graphid().tileid(), tile->header()->graphid().level(), edge - start_edge);
     for(const auto& i : intersection) {
-      //to avoid dups and minimize having to leave the tile for shape we:
-      //always write a given edge to the tile it originates in
-      bool originating = i.first == start_id;
-      //never write a given edge to the tile it terminates in
-      bool terminating = i.first == end_id;
-      //write a given edge to intermediate tiles only if its originating tile id is < its terminating tile id
-      bool intermediate = i.first < end_id;
```
kevinkreiser: this was wrong! it should have been `start_id < end_id`

src/mjolnir/graphtilebuilder.cc
```diff
@@ -721,33 +721,33 @@ std::array<std::vector<GraphId>, kBinCount> GraphTileBuilder::BinEdges(const Til
...
+    //writing the edge to the tile it originates in
+    //not writing the edge to the tile it terminates in
+    //writing the edge to tweeners if originating < terminating
     auto start_id = tiles.TileId(edge->forward() ? shape.front() : shape.back());
     auto end_id = tiles.TileId(edge->forward() ? shape.back() : shape.front());
+    auto intermediate = start_id < end_id;
 
-    //bin some in, save some for later, ignore some
+    //if this starts and ends in the same tile and we've seen it already we can skip it
+    if(start_id == end_id && !ids.insert(edge->edgeinfo_offset()).second)
```
kevinkreiser: just having seen the edgeinfo offset before isnt enough when you are dealing with larger tiles binning into smaller ones. you need to know where they exist with respect to the smaller tiles.

gknisely:
:ship:

noblige:
awesome! will try it out!

PR #567, opened 2016-11-22
Speed limits

gknisely:
@dnesbitt61 let us know if this will all still makes sense in the costing logic.

speed = average speed if exists, else advisory speed if exists, else max_speed if exists, else categorized speed. 

speed_limit = max_speed

I left the tagged_speed flag for now.  Do you think you will need it in the costing?


gknisely:
depends on baldr updates

dnesbitt61:
:shipit:

PR #571, opened 2016-11-30
Update test files with end_offset now fixed at the end of the header.

dnesbitt61:
depends on baldr change


gknisely:
:+1: after you rerun circle and checks pass

PR #574, opened 2016-12-03
actually set speed limit

noblige:


gknisely:
good catch.  Thank you.

PR #579, opened 2016-12-13
Cr issue

gknisely:
added logic to stop infinite loop on bad restrictions.
added some more stats.
Tested with planet.


dgearhart:
:shipit:

PR #583, opened 2017-01-04
Support for adding traffic segment associations to a tile.

dnesbitt61:
This just includes code to add information to the tile and to the tile header. The program to actually associate traffic to Valhalla edges is not contained within mjolnir since it has additional dependencies.


kevinkreiser:
:+1:

PR #584, opened 2017-01-05
removed Paris tz hack.

gknisely:
removed Paris hack.  Fallback to tz db if tz string not found in region list.  Added support for suburban_railway.


dgearhart:
:shipit:

PR #594, opened 2017-01-20
Modify GraphValidator to handle tiles which have no admin entries.

zerebubuth:
This occurs in some of the test tiles I've been creating. It looks like the rest of the code can handle `endnodeiso` and `begin_node_iso` being empty strings, it's just the `tile->admin(0)` lookup that causes the problems when there are no admin entries.

@kevinkreiser could you take a look, please?

### DIFF
--- a/src/mjolnir/graphvalidator.cc
+++ b/src/mjolnir/graphvalidator.cc
@@ -91,7 +91,11 @@ uint32_t GetOpposingEdgeIndex(const GraphId& startnode, DirectedEdge& edge,
   bool sametile = (startnode.tileid() == endnode.tileid());
 
   // Set the end node iso.  Used for country crossings.
-  endnodeiso = end_tile->admin(nodeinfo->admin_index())->country_iso();
+  if (end_tile->header()->admincount() > 0) {
+    endnodeiso = end_tile->admin(nodeinfo->admin_index())->country_iso();
+  } else {
+    endnodeiso.clear();
+  }
 
   // Set the deadend flag (TODO - do we need to worry about driveability)
   deadend = nodeinfo->intersection() == IntersectionType::kDeadEnd;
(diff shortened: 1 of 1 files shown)


kevinkreiser:
this looks fine to me. i have to admit i dont understand why we havent run into this before though. id ship this but i would like @gknisely to take a look if he had a second

gknisely:
Why is this a problem?  Or I guess what issue are you seeing?

If there is no admin set, the admin_index() should be zero and a blank or empty admin should always exist in every tile at index 0.

gknisely:
For example://  In graphbuilder...

      // Create a dummy admin record at index 0. Used if admin records
      // are not used/created or if none is found.
      graphtile.AddAdmin("None","None","","");

dnesbitt61:
Are these tests with tiles you are creating? If so maybe you missed adding the dummy admin record as Greg's code sample pointed out?

kevinkreiser:
@gknisely thanks for that, i knew that was somewhere but i just couldnt find it! essentially if you are making fake tiles (ie for writing unit tests) you need to do that step otherwise the code above will throw an exception without @zerebubuth 's patch above. maybe a `graphtilebuilder` could in the case its making a new tile just do this step so you dont have to know to do it manually. or maybe that has other issues.

zerebubuth:
In this case, I was seeing a problem building a tile which didn't automatically get a zero admin set.

I'll add the `AddAdmin` code to my test builder, but I think it's worthwhile either making this not a problem any more (as this patch does), or asserting that there's a zero admin in each tile (e.g: in `GraphTile::StoreTileData` or similar). Hopefully this will save anyone else who tries to build tiles other than from an OSM PBF file a few hours of debugging :smile: 

What do you think?

gknisely:
I vote for asserting....what does everyone else think?

kevinkreiser:
i would vote to make `graphtilebuilder` handle it automatically. when you go to write the tile it would just always write the `null` admin followed by however many you've told it to write. forcing people to remember to do this is a mental burden we dont need if we can avoid it. maybe im oversimplifying though and its not as easy as i suggest

dnesbitt61:
I would add it to the GraphTileBuilder constructor. If it encounters a new/empty tile add the empty admin record. If I remember we require this to be the first admin record so we should force it to be added so any subsequent AddAdmin calls work properly.

kevinkreiser:
im going to close this and add an issue to handle this as dave suggested making sure tiles always have it

PR #601, opened 2017-01-29
Update to use new TrafficAssociation and TrafficChunk structures

dnesbitt61:
Update interface for adding traffic association. Separate into 2 methods: AddTrafficSegment for edges that associate to a single traffic segment and AddTrafficSegments for edges that associate to more than 1 traffic segment.


Makefile.am
```diff
@@ -89,8 +89,7 @@ nobase_include_HEADERS = \
 	valhalla/mjolnir/restrictionbuilder.h \
 	valhalla/mjolnir/shortcutbuilder.h \
 	valhalla/mjolnir/transitbuilder.h \
-	valhalla/mjolnir/util.h \
-	valhalla/mjolnir/validatetransit.h
```
kevinkreiser: this looks like  a mistake

scripts/dependencies.sh
```diff
@@ -10,7 +10,7 @@ if [[ $(grep -cF trusty /etc/lsb-release) > 0 ]]; then
 fi
 sudo add-apt-repository -y ppa:kevinkreiser/prime-server
 sudo apt-get update
-sudo apt-get install -y autoconf automake pkg-config libtool make pkg-config gcc g++ lcov vim-common libboost1.54-all-dev protobuf-compiler libprotobuf-dev lua5.2 liblua5.2-dev libsqlite3-dev libspatialite-dev libgeos-dev libgeos++-dev libcurl4-openssl-dev libprime-server0.6.3-dev
```
kevinkreiser: this is also a mistake

src/mjolnir/valhalla_build_transit.cc
```diff
@@ -14,7 +14,6 @@
 #include <boost/format.hpp>
 #include <boost/filesystem/operations.hpp>
 #include <boost/algorithm/string.hpp>
-#include <boost/tokenizer.hpp>
```
kevinkreiser: mistake

PR #602, opened 2017-01-30
Update argument processing

### LINKED ISSUE #599: Don't claim missing --config is a bug
```bash
root@6e71ed10e3ce:/data# valhalla_build_tiles
Unable to parse command line options because: the option '--config' is required but missing
This is a bug, please report it at https://github.com/valhalla/mjolnir/issues
root@6e71ed10e3ce:/data#                               
```

I don't think _This is a bug_ is true, but I'm guessing it's part of a default message when an exception is handled. Should this be removed for this case, where it's the user's fault?

### LINKED ISSUE #600: Exit when there are no files provided
```
root@6e71ed10e3ce:/data# valhalla_build_tiles --config /source/conf/valhalla.json 
2017/01/28 04:00:37.183004 [WARN] Non-empty /data/valhalla/2 will be purged of tiles
2017/01/28 04:00:39.402706 [INFO] Parsing files: 
2017/01/28 04:00:39.403241 [INFO] Parsing ways...
[ attempts to run ]
```

If the user doesn't provide any valid files to parse, Valhalla should probably (at least) warn or (maybe better) exit with an error. Right now it looks like a lot happened, but really all it did was nothing from what I can tell. Is this a real use case?

dnesbitt61:
Refer to app. as valhalla_build_tiles consistently. Update argument parsing to allow -h and/or -c by themselves. Update error message for missing config file and check if no input files are specified.

Fixes #599 and fixes #600

### DIFF
--- a/src/mjolnir/valhalla_build_tiles.cc
+++ b/src/mjolnir/valhalla_build_tiles.cc
@@ -59,27 +59,23 @@ int main(int argc, char** argv) {
     return EXIT_FAILURE;
   }
 
-  bool info_only = false;
+  // Print out help or version and return
   if (vm.count("help")) {
     std::cout << options << "\n";
-    info_only = true;
+    return EXIT_SUCCESS;
   }
   if (vm.count("version")) {
     std::cout << "valhalla_build_tiles " << VERSION << "\n";
-    info_only = true;
+    return EXIT_SUCCESS;
   }
 
   // Exit if no config file or no input pbf files
   if (!vm.count("config") || !boost::filesystem::is_regular_file(config_file_path)) {
-    if (!info_only) {
-      std::cerr << "Configuration file is required\n\n" << options << "\n\n";
-    }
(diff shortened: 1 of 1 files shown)


src/mjolnir/valhalla_build_tiles.cc
```diff
@@ -28,73 +28,62 @@ using namespace valhalla::mjolnir;
...
-    return false;
+    std::cerr << "Unable to parse command line options because: " << e.what() << "\n";
+    return EXIT_FAILURE;
   }
 
+  bool info_only = false;
   if (vm.count("help")) {
     std::cout << options << "\n";
-    return true;
+    info_only = true;
```
kevinkreiser: why not just return `EXIT_SUCCESS` if they typed `--help`? do we need to validate the extra arguments if they are asking for help?

kevinkreiser:
change if you feel like it, if not ship it

PR #603, opened 2017-01-30
added url encoding 

gknisely:
added url encoding to one_stopids and route_ids.  Also, added param to exclude_geometry for routes.  We use the geometry from the route stop patterns.


dgearhart:
:shipit:

PR #604, opened 2017-01-30
Update to separate into AddTrafficSegment and AddTrafficSegments.

dnesbitt61:
Use new TrafficAssociation and TrafficChunk structures.


kevinkreiser:
:rocket:

